PR #11, opened 2021-12-19
Remove trailing comma in compile_commands.json

JKurland:
This removes the trailing comma in the output, even though clangd doesn't care about the comma, the c/c++ extension for vscode does. I currently get better results from the default c/c++ extension than from clangd (probably because my actual build uses gcc).

I haven't don't a huge amount of testing, just checked that it works for my project, I'm just running `bazel run @hedron_compile_commands//:refresh_all`

### DIFF
--- a/extract.py
+++ b/extract.py
@@ -204,12 +204,16 @@ if __name__ == '__main__':
     bazel_workspace_dir = os.environ['BUILD_WORKSPACE_DIRECTORY'] # Set by `bazel run`. Can't call `bazel info workspace` because bazel is running us outside the workspace.
     # Bazel gotcha warning: If you were tempted to use `bazel info execution_root` as the build working directory for compile_commands...search ImplementationReadme.md to learn why that breaks.
 
-    # Dump em to stdout as compile_commands.json entries
+    # Dump em to stdout as compile_commands.json 
+    first = True
     for files, command in outputs:
         for file in files:
+            if not first:
+                sys.stdout.write(',')
+            else:
+                first = False
             sys.stdout.write(json.dumps({
                 'file': file,
                 'command': command,
                 'directory': bazel_workspace_dir
             }, indent=2, check_circular=False))
-            sys.stdout.write(',')
--- a/extract.py
+++ b/extract.py
@@ -204,16 +204,12 @@ if __name__ == '__main__':
     bazel_workspace_dir = os.environ['BUILD_WORKSPACE_DIRECTORY'] # Set by `bazel run`. Can't call `bazel info workspace` because bazel is running us outside the workspace.
     # Bazel gotcha warning: If you were tempted to use `bazel info execution_root` as the build working directory for compile_commands...search ImplementationReadme.md to learn why that breaks.
 
-    # Dump em to stdout as compile_commands.json 
-    first = True
+    # Dump em to stdout as compile_commands.json entries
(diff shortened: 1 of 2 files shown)


cpsauer:
Hey, @JKurland! Thanks for giving the plugin a try, finding an issue, and contributing. Sorry that comma gave trouble.

Unless I've missed something, I think this change would removes the comma at the end of the output for *each* Bazel target, rather than just the one at the end of the list. How about if I play around with things a little bit to make sure it's fixed?

cpsauer:
Great! So there's a quick fix with sed. Could you confirm that works for you? (Have tested and seems to work well locally.)

cpsauer:
And before you go, @JKurland, I'd love to ask a couple questions so I can help you and others better in the future.

1. I'm curious in which ways you found the default extension C++ extension better than clangd!
     Doubly because we'd found clangd better when we last checked in, but that was a couple months ago. I'd especially like to know if there's any chance we were messing up something with gcc<>clangd. Were clangd features clearly working properly? And if not, any chance you'd take a peek for clangd errors in the log at [Terminal Tab]>Output>[Dropdown]>clangd?
2. If you're down to share, I'd also love to hear how it's working for you otherwise--and what you're using it for. Helps me know how to make it better and understand it's usefulness. And if there are any other rough edges you see, I'd love to hear about them!

-Chris

JKurland:
Ah yeah, that makes sense about the multiple targets, thanks.

1. Clangd was just giving me a lot of false errors (not that c/c++ extension doesn't) but clangd was giving more. It seemed to have real trouble with member initialisation lists. That might have just been a configuration problem on my end but I'm also using coroutines which it was also having trouble with and I assumed that was because clang doesn't support coroutines yet. For example it didn't understand the -fcoroutine flag that gcc requires.

2.I'm just using it personally, I used the grailbio compdb generator before too but just wanted to give this a try, especially since the grailbio one had the nasty feature that it wants you to give it the execroot during the bazel build. Even if it wanted the workspace root instead it still gets in the way of the hermeticness. So using bazel run instead of bazel build to generate the database makes sense. I did prefer how in grailbio "build ..." would also update the compdb, but that's not a big deal but some way to automate rebuilding the database would be cool, can't think of much except for rerunning on a schedule though.

JKurland:
This doesn't work for me I'm afraid, i have to change
`sed -i '' '$ s/,$//' compile_commands.json` to 
`sed -i '$ s/,$//' compile_commands.json`
otherwise i get and error
`sed: can't read $ s/,$//: No such file or directory`
and
```
$ cat compile_commands.json | jq
parse error: Unfinished JSON term at EOF at line 131949, column 2
```
which is the ending comma and no "]"

cpsauer:
Oh no! Okay--probably a Linux vs BSD sed difference. I'm guessing we need a -e before the sed program.

Will fix tm--it's super late here.

cpsauer:
Couldn't help myself and stayed up. Does that fix things?
(Booted myself a linux VM to check. Guessed rightly that it was a GNU<>BSD sed difference, but a slightly uglier one that I'd expected. Anyway, that worked in both my tests.)

Thanks so much for taking the time to reply on the other questions, btw. Really appreciate it.

JKurland:
No worries, thanks for fixing it.
Yeah that works for me.

PR #17, opened 2022-01-14
fixup expected name of the local repository

lummax:
As this lives in a git repository called `bazel-compile-commands-extractor`, I guess it
is save to assume, that the local checkout is named the same. So lets make the hint on
how to setup a local developent environment less confusing.


cpsauer:
Great catch. Thank you, @lummax! I really appreciate your improvements.

[My sincere apologies for being slow in getting to these. Last week got crazy, and I couldn't respond as fast as I would have liked. I'll get back to you on the other one tomorrow.]

Issue #8, opened 2021-11-25
Windows!


cpsauer:
Update: Windows support landed! Thanks all. The latest commits should have it. You can go ahead and start using this tool on Windows.

We haven't yet tested on Windows. Windows might need some patching parallel to that for macOS (in [refresh.template.py](https://github.com/hedronvision/bazel-compile-commands-extractor/blob/main/refresh.template.py)), but it should be a relatively easy adaptation compared to writing things from scratch. If you're trying to use it on Windows, let us know here. We'd love to work together to get things working smoothly.

lummax:
Just to leave a quick note while I am researching this:

- it seems we should not just do `command = ' '.join(args)` as the path to `cl.exe` contains whitespaces. I guess we could use https://stackoverflow.com/a/12130261
- `/showIncludes` and `/P` are the flags we are interested here
  - https://docs.microsoft.com/en-us/cpp/build/reference/showincludes-list-include-files?view=msvc-160
  - https://docs.microsoft.com/en-us/cpp/build/reference/p-preprocess-to-a-file?view=msvc-170
- I get a lot of warnings about missing header files from the C++ stdlib. https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1083?view=msvc-170#the-command-line-include-or-lib-environment-is-not-set is a good pointe and indeed the bazel default toolchain sets these environment variables https://github.com/bazelbuild/bazel/blob/6dc941e58dfc1d4a9714a76b921fbe11fce658ed/tools/cpp/windows_cc_toolchain_config.bzl#L627. Currently I am not sure if the `bazel aquery` exposes that correctly. For some other action I see `environmentVariables` but not for the compile actions.

cpsauer:
Hi, @lummax! Thanks for diving in here. Looks like good stuff to me. Tapping back some thoughts, in case they're helpful.
- Yep, I definitely should have considered shell escaping. Good call.
  - Alternatively, we could use the `arguments` array option in [the compile_commands.json spec](https://clang.llvm.org/docs/JSONCompilationDatabase.html)?
- 👍🏻
[...]

PR #25, opened 2022-02-07
add support on Windows

LoSealL:
> Issue related: #8 
To make things happen on Windows, we need a batch template which do simillar stuffs as the shell template.
And I try to escape from generating headers from CL which seems not quite feasible. But **clangd** managed to work without the header files as long as the `/I` is correct for searching headers.

### DIFF
--- a/extract.py
+++ b/extract.py
@@ -38,6 +38,11 @@ def _get_headers(compile_args: List[str], source_path_for_sanity_check: Optional
 
     Relatively slow. Requires running the C preprocessor.
     """
+    if os.name == 'nt':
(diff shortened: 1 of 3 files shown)


cpsauer:
[Supplementing https://github.com/hedronvision/bazel-compile-commands-extractor/issues/8#issuecomment-1032301412]

@LoSealL, I super appreciate your efforts--and closing out PRs makes me sad--but I think I'd better ask you to fold your efforts into his. I wish we'd gotten you aware of what he was working on before you started up!

If you'd like me to reopen or think we should change course, just say, and I'll open this back up.

PR #27, opened 2022-02-10
Python3.7 compatibility

lummax:


lummax:
I am sorry, there is no option to allow maintainer edits.

refresh.template.py
```diff
@@ -309,7 +308,7 @@ def _get_commands(target: str, flags: str):
         # object_hook allows object.member syntax, just like a proto, while avoiding the protobuf dependency
         parsed_aquery_output = json.loads(aquery_process.stdout, object_hook=lambda d: types.SimpleNamespace(**d))
     except json.JSONDecodeError:
-        print("aquery failed. Command:", shlex.join(cmd), file=sys.stderr)
+        print("aquery failed. Command:", subprocess.list2cmdline(cmd), file=sys.stderr)
```
matta, line 317: Is it a good idea to call `subprocess.list2cmdline()`?  From this comment https://github.com/python/cpython/blob/14284b0e71c2110f94941b774478fb594d0ef192/Lib/subprocess.py#L65-L66 and https://bugs.python.org/issue10838 it looks like the Python maintainers intend for it to be a private API.

How about this instead?

```python
' '.join(shlex.quote(arg) for arg in cmd)`
```

lummax:
> Any chance we could get you to just upgrade your python instead? 

Unfortunately not, we will keep Python 3.7 as the min-version for some time.

> Offhand, I'd be inclined towards the command string over arg list for human readability and use in debugging, but I could be convinced.

 I do understand it and i do concur. On the other had, `bazel` gives us a list so why be clever and potentially introduce escaping/whitespace errors?

> But giving me commit access to the fork would work! 

Unfortunately, this is not a decision I can make.

refresh.template.py
lummax: Good catch. Generally I prefer the `subprocess.list2cmdline()` even though that is internal API as I know that is what is used internally and should cope with escaping problems. But this is just for displaying the string to the user. I'll change it.

cpsauer:
Ok! For my learning, any particular reason I should know keeping you guys (and maybe others) on 3.7?

Bummer on access; would speed these up. Maybe there's another creative solution, like a personal fork?

I'll merge in a sec, with some cleanups (easier to just do than go around, but if you disagree, just toss up a PR).

For now, I'm going to keep using the command string, for the site same human-readability reasons as above, with our backwards-compatible escaping. But if you guys see any issues pop up, especially on Windows, please say something and we'll switch.

cpsauer:
Friendly bump on questions, above!

lummax:
Sorry :)

> For my learning, any particular reason I should know keeping you guys (and maybe others) on 3.7?

Not really. We are shipping a product with plugin support via python and some customers are using Python 3.7, so we are using that as the min-version.

> But if you guys see any issues pop up, especially on Windows, please say something and we'll switch over to the args list.

It seems to work on the other PR, let's see :)

cpsauer:
Got it! Though now I'm curious what you're building with this :)

At some point, folks may want to just use a more modern python version, though I'm super happy to stay backwards compatible if it's needed. Could I ask you to also encourage your users to update, since it's super easy (in my experience, though again, maybe there's some good reason), and that version is getting up towards 4 years old? It'd benefit them quite a bit more than it'd benefit us, I bet :)

PR #28, opened 2022-02-14
refresh.py: basic windows support

lummax:


refresh.template.py
```diff
@@ -90,6 +97,57 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
...
+            (
+                os.fspath(base / "ATLMFC/include"),
+                os.fspath(base / "include"),
+                os.fspath(windows_kits / "ucrt"),
+                os.fspath(windows_kits / "shared"),
+                os.fspath(windows_kits / "um"),
+                os.fspath(windows_kits / "winrt"),
+                os.fspath(windows_kits / "cppwinrt"),
+            )
+        )
```
cpsauer, line 126: Phew. I remember your saying that it needed this. Thanks for finding a way through.
[I'm assuming this means that Bazel didn't pass the environment variable in aquery? I think that's worth filing a Bazel issue about.]

Questions that occur to me, not knowing that much about Windows: 
- Will `base` work with clang-cl?
- Will the `10` path work on other versions of Windows?

refresh.template.py
```diff
@@ -90,6 +97,57 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
...
+    if not header_search_process.stderr:
+        return [] # Worst case, we couldn't get the headers,
+
+    headers = (
+        # TODO This is not enough. Based on the locale, `cl.exe` will emit different marker strings. This should be extendend to multiple languages.
+        line.replace("Note: including file:", "").strip()
+        for line in header_search_process.stderr.splitlines()
+        if "Note: including file:" in line
+    )
+
```
cpsauer, line 147: Worth also printing any real errors?

refresh.template.py
```diff
@@ -90,6 +97,57 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
...
+        env=environ,
+        encoding="utf-8",
+        check=False, # We explicitly ignore errors and carry on.
+    )
+
+    if not header_search_process.stderr:
+        return [] # Worst case, we couldn't get the headers,
+
+    headers = (
+        # TODO This is not enough. Based on the locale, `cl.exe` will emit different marker strings. This should be extendend to multiple languages.
```
cpsauer, line 142: Ugh, I'm so sorry MSVC does this.

Brainstorming some crazy ideas:
- We could try to call Bazel's code? (Not sure why they're extracting headers, but I remember that code you linked.)
- Is there a way to set the compiler's locale?
- We could invoke the compiler twice, with and without /showIncludes, and scan the lines that differ for the last instance of ": "?

refresh.template.py
```diff
@@ -90,6 +97,57 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
...
+        header_cmd,
+        cwd=os.environ["BUILD_WORKSPACE_DIRECTORY"],
+        stderr=subprocess.PIPE,
+        stdout=subprocess.DEVNULL,
+        env=environ,
+        encoding="utf-8",
+        check=False, # We explicitly ignore errors and carry on.
+    )
+
+    if not header_search_process.stderr:
```
cpsauer, line 138: Protecting against None? Otherwise handled already, right?

refresh.template.py
```diff
@@ -90,6 +97,57 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
...
+        return [] # Worst case, we couldn't get the headers,
+
+    headers = (
+        # TODO This is not enough. Based on the locale, `cl.exe` will emit different marker strings. This should be extendend to multiple languages.
+        line.replace("Note: including file:", "").strip()
+        for line in header_search_process.stderr.splitlines()
+        if "Note: including file:" in line
+    )
+
+    return list(set(headers))
```
cpsauer, line 148: Oh no. Do they also have duplicates?

refresh.template.py
```diff
@@ -221,12 +279,15 @@ def _all_platform_patch(compile_args: typing.List[str]):
 
     return list(compile_args)
 
+def _windows_platform_patch(compile_args: typing.List[str]):
+    if platform.system() == "Windows":
+        return list(arg[0] + arg[1:].replace("/", "\\") for arg in compile_args)
+    return compile_args
```
cpsauer, line 285: It's not obvious to me what this does!

cpsauer (commented):
Yay! Thank you, Lukas. Super exciting to bring this to Windows :) I really appreciate your great work.

So I know (and in the absence of a description): Is there more to come here/is this a WIP? Or should I be looking to merge? 

In the meantime, I'll read through and write down some quick comments.

[Also, I don't have a Windows machine handy at the moment so I'm going to have to ask you to be responsible for making sure it works. But it sounds like you're already running/using some variant of this.]

Thanks again!
Chris

lummax:
> So I know (and in the absence of a description): Is there more to come here/is this a WIP? Or should I be looking to merge?

This is working as-is for us. But this is "basic" as there are still open points about multi-language/encoding support. I can't really answer them.

One option for the multi-language support would be to research and include all the marker strings. Don't know if we want that or if we wait for users to pop in and add support one-by-one (and provide a great opportunity for a good first PR :D)

> [Also, I don't have a Windows machine handy at the moment so I'm going to have to ask you to be responsible for making sure it works. But it sounds like you're already running/using some variant of this.]


Yes.

refresh.template.py
```diff
@@ -50,6 +51,12 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
     # As an alternative approach, you might consider trying to get the headers by inspecing the Middlemen actions in the aquery output, but I don't see a way to get just the ones actually #included--or an easy way to get the system headers--without invoking the preprocessor's header search logic.
         # For more on this, see https://github.com/hedronvision/bazel-compile-commands-extractor/issues/5#issuecomment-1031148373
 
+    if "cl.exe" in compile_args[0]:
+        return _get_headers_msvc(compile_args)
+    return _get_headers_gnu(compile_args, source_path_for_sanity_check)
+
+
+def _get_headers_gnu(compile_args: typing.List[str], source_path_for_sanity_check: typing.Optional[str] = None):
```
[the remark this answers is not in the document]
lummax: Sorry, I was unable to infer the right order of functions.
gcc vs. gnu: happy with both and I don't have a strong opition.

Will change it :)

refresh.template.py
```diff
@@ -90,6 +97,57 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
     return headers
 
 
+def _get_headers_msvc(compile_args: typing.List[str]):
+    # We don't want to compile, but only pre-process and show all the included header files.
```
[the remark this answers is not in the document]
lummax: Sorry, I am still in the habit of aligning everything to 80 chars.

refresh.template.py
```diff
@@ -221,12 +279,15 @@ def _all_platform_patch(compile_args: typing.List[str]):
 
     return list(compile_args)
 
+def _windows_platform_patch(compile_args: typing.List[str]):
```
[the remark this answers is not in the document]
lummax: We use `black` to format our python code. So I did get a bit lax on trying to match existing styles.

refresh.template.py
lummax: I don't know the answers to these questions. I am a linux guy. I just have to support some developers on windows.

refresh.template.py
lummax: I am not sure how to detect real errors. Because `cl.exe` prints the pre-processed file to `stderr`. But also some error-lines. Will look into it.

refresh.template.py
lummax: > We could try to call Bazel's code? (Not sure why they're extracting headers, but I remember that code you linked.)

That is Java code, so I guess not.

> Is there a way to set the compiler's locale?

No idea. Might depend on installed language packs for MSVC.

> We could invoke the compiler twice, with and without /showIncludes, and scan the lines that differ for the last instance of ": "?

Would increase generation time for the `compile_commands.json` quite drastically.

refresh.template.py
lummax: You are right, there is no need.

refresh.template.py
lummax: I am actually not sure. I cargo-culted this :D

refresh.template.py
lummax: Not sure why I did the `arg[0] + arg[1:]` dance. Might have been an experiment. Removed.

refresh.template.py
lummax: Done.

lummax:
Sorry, the force-push hides the fix-up commits. There are several accessible from https://github.com/hedronvision/bazel-compile-commands-extractor/commit/5252003463828444d439ff663065d8d72c5a4175

refresh.template.py
```diff
@@ -50,6 +51,12 @@ def _get_headers(compile_args: typing.List[str], source_path_for_sanity_check: t
     # As an alternative approach, you might consider trying to get the headers by inspecing the Middlemen actions in the aquery output, but I don't see a way to get just the ones actually #included--or an easy way to get the system headers--without invoking the preprocessor's header search logic.
         # For more on this, see https://github.com/hedronvision/bazel-compile-commands-extractor/issues/5#issuecomment-1031148373
 
+    if "cl.exe" in compile_args[0]:
```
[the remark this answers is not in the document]
cpsauer: Could you check the clang toolchain you'd posted in that other issue, and commenting what you find?
(Great to hear no wrappers for msvc.)

Also: thoughts on using `endswith`?

refresh.template.py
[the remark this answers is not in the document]
cpsauer: No, no. Sorry to be wasting time on these minor things. Again normally I'd just propose changes. But in the absence of having permissions I think should still say something, since we're working together for a while here.

refresh.template.py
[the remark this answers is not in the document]
cpsauer: Ideally above _all_platforms_patch, but I can do on merge if you'd rather.

refresh.template.py
cpsauer: How about just printing all the lines that don't match as starting with "Note: including file:"?

...at least if there's not too much other logging? I'd just like to prevent silent failures for people, especially given msvc being hard to parse. Totally get that it's annoying to parse both out of stderr.

Also, maybe `startswith`?

refresh.template.py
cpsauer: I'd really like it if we could make this more robust somehow

refresh.template.py
cpsauer: heh good phrase

But might be worth a quick check? I *think* they'll list it each time its included, but I'm not sure.

refresh.template.py
cpsauer: Oh, wow. Thanks. I'd totally misunderstood this the first time.
Does it indeed break without this? (In my memory, Windows started supporting / in 7, and Bazel doesn't support older than that. But I could be wrong.)

cpsauer:
Not a problem on the force pushes. This is all small enough that it's easy for me to scan through. Thanks for having kept things incremental!

I would really like to ask for thing things to be a bit more robust before we merge, especially because my ability to support Windows folks well is more limited. Tapped back some comments above. Thanks for your patience on these :)

refresh.template.py
cpsauer: I realized also: Should we be detecting a windows command via the compiler name, as you did above, rather than detecting the host platform? Might future proof us better for, for example, cross-compilation and remote builds.

refresh.template.py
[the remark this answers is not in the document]
lummax: https://github.com/grailbio/bazel-toolchain/issues/4 there is no support for Windows yet. :(

`endswith()`: good catch :)

refresh.template.py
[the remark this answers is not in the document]
lummax: Done.

refresh.template.py
lummax: I think I saw some stuff during development, but I am unable to create a reproducer that produces a lot of spam. So let's try this for now and see how that behaves.

refresh.template.py
lummax: Ok, could not see duplicates. Removed.

refresh.template.py
lummax: > Does it indeed break without this?

Removed and still works. I might have introduce it because the `/` also end up in the `compile_commands.json`. But `clangd` seems to cope with this aswell. Removing this.

refresh.template.py
lummax: > I did do some more digging, and maybe setting VSLANG=1033 would do the trick, getting everything always printed in english? (I saw https://github.com/ninja-build/ninja/issues/613) If it works, that would be so much better.

I tried setting it to turkish just for the header finding `cl.exe` call using this trick and it did not work. A comment says "before calling vcvars32/64" and that is out of our control.

lummax: > If that fails, at a minimum, could we match the language prefixes that Bazel does and you linked to? (Languages also listed, maybe more easily, in that ninja issue.

Done.

cpsauer:
About to head to sleep over here, but I got curious when I saw your comment about vcvars, and started reading up

Looks like there might be something very interesting there that could solve multiple problems for us. In addition to language, it looks like that command is what's supposed to be responsible for setting up the required environment variables, like INCLUDE. Running it in the script might solve both problems!

Some quick links, to hand things off in our cross-timezone relay:
- A general, helpful article: https://renenyffenegger.ch/notes/Windows/development/Visual-Studio/environment-variables/index
- Microsoft docs: https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170
- Bazel calls to it that you might be able to draw on: https://github.com/bazelbuild/bazel/blob/0ba4caa5fc4bfd2d7f9af2a1653ef5b4bd83d176/tools/cpp/windows_cc_configure.bzl

refresh.template.py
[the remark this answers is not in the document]
cpsauer: Whoops. Sorry. That's on me.
[I'd searched windows in their codebase, seen some references, and wrongly concluded there was support.]

Minor, but any chance we could add a comment mentioning that we're also trying to catch clang-cl with the endswith?
(I can also do on merge, but writing this as a note, rather than resolving, to make sure that collectively we don't forget.)

refresh.template.py
cpsauer: Awesome. Thanks for your care and double checking.

refresh.template.py
cpsauer: Super good for me to know--and great that Window's and clangd's forward path support eliminates the need to handle this case. Thank you!

cpsauer:
@lummax, just checking in to make sure I'm not holding things up here. Super excited to merge this, and we're almost there! 
But I'd love to ask for your help investigating the vcvars stuff (above) on your windows machine, in the hope that it might radically simplify things

cpsauer:
Noting down to either you or future me that we also want to update the README as part of this change. Decidedly awesome windows support, rather than basic, I'd say.

refresh.template.py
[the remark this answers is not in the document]
lummax: Done and rebased.

lummax:
Sorry for being unresponsive. I have to squeeze working on this into the other obligations.

> investigating the vcvars stuff 

I see three ways that I both don't like:

- call that vsvarsall.bat once and extractinng the required environment
- create a wrapper that invokes every compilation action inside the correct environment by calling vsvarsall.bat for each one
- detect Windows at the very start of the `extract.py` and somehow call the script again but under an environment where vsvarsall.bat has run

I don't like any of them and would rather depend on the user to have that script called in their shell environment.

> update the README as part of this change
I'll leave that up to you :)

cpsauer:
Totally get it, and I appreciate your help. 

[Same situation over here, so I empathize! :) Sorry to be pushing for quality and not coding. Worst case I can spin up a Windows VM and dive in; however, I would really appreciate your finishing this out if you can, since you have things set up.]

Re the vcvars options, wouldn't any of those three better than the current hardcoding, if they work? That said, there might be even easier options. How about: 
(1) For the language settings: what about doing a manual test run of vcvarsall.bat with and without the VSLANG setting, inspecting the environment to see how vcvars is setting cl.exe to English (assuming that does indeed work, might also need [chcp 437](https://stackoverflow.com/questions/2286216/how-to-change-msbuild-error-message-language)?)). Then just directly set whatever environment variable vcvarsall uses in the python subprocess call.  
(2) For INCLUDE: Looks like you might be able to get Bazel to give it to you via starlark by calling in through the toolchain (or through the link above).

cpsauer:
@lummax, could I ask you to chase down those last couple leads so we can get this merged in? 
Would be pretty awesome to know that we're as robust and clean as can be.

lummax:
> (1) For the language settings: what about doing a manual test run of vcvarsall.bat with and without the VSLANG setting, inspecting the environment to see how vcvars is setting cl.exe to English (assuming that does indeed work, might also need [chcp 437](https://stackoverflow.com/questions/2286216/how-to-change-msbuild-error-message-language)?)). Then just directly set whatever environment variable vcvarsall uses in the python subprocess call.

I can't see revelant changes when setting `VSLANG` and re-running `vcvarsall.bat`.

I am skeptical of using anything related to the `vcvarsall.bat` because it is not-trivial to find it. `bazel` does for the autoconfigure:

- locate a `vswhere.exe`: https://github.com/bazelbuild/bazel/blob/efb2b80953983dce499d453a9f55a74ffaf8c42d/tools/cpp/windows_cc_configure.bzl#L159
- generate a bat-file to run to extract more version information: https://github.com/bazelbuild/bazel/blob/efb2b80953983dce499d453a9f55a74ffaf8c42d/tools/cpp/windows_cc_configure.bzl#L159
- fall back to the registry if both steps above fail: https://github.com/bazelbuild/bazel/blob/efb2b80953983dce499d453a9f55a74ffaf8c42d/tools/cpp/windows_cc_configure.bzl#L213
- locate a `vcvarsall.bat` depending on the VS version: https://github.com/bazelbuild/bazel/blob/efb2b80953983dce499d453a9f55a74ffaf8c42d/tools/cpp/windows_cc_configure.bzl#L276
- call the `vcvarsall.bat` with specific arguments: https://github.com/bazelbuild/bazel/blob/efb2b80953983dce499d453a9f55a74ffaf8c42d/tools/cpp/windows_cc_configure.bzl#L378

So this is already hard to find the correct `vcvarsall.bat` and to do it in a robust fashion. And configurable with some environment variables: https://bazel.build/docs/windows#build_cpp. This is just the autoconfigure of the bazel CC toolchain. The user is free to setup their own CC toolchain.

I get the wish to make it as robust as possible. But IMHO this is a little out of the scope of the `bazel-compile-commands-extractor` if the user can get the same result by calling `vcvarsall.bat` themselves or setting up the right environment variables.

cpsauer:
Wahoo! We're merged--in good final form https://github.com/hedronvision/bazel-compile-commands-extractor/commit/3b14e77f97d4a2a596a017f069b9641a07919698. Many thanks, @lummax for laying the foundation of Windows support!

In retrospect, I made a rookie maintainer mistake by taking not merging this *way* earlier, and having us go incrementally from there. I also somewhat butchered git history dealing with the Windows \r\n. [Enough that GitHub didn't tie the commit to this PR, so I'll have to manually close.]  I hope you'll forgive me.

Regardless, all your code is in, with many thanks.

If you're curious how some of the above resolved: 
- Turns out setting the compiler language via VSLANG does work...it just it doesn't work if you don't have the target language pack installed. Since we can't guarantee people will have English, that makes your hardcode-all-languages approach the cleanest we've got, I think. I added a note to that effect--and also wrote that to the Ninja folks 
- It was indeed possible to get the results to INCLUDE from the (top-level) Bazel toolchain. They really should be providing it in aquery, though. Updated [my macOS issue to include that](https://github.com/bazelbuild/bazel/issues/12852).

PR #29, opened 2022-02-14
refresh.py: cope with no action/empty `aquery` JSON

lummax:


cpsauer:
Good catch! Definitely an edge case I'd missed in the switch to JSON from protos.

cpsauer:
[But after some quick testing, I think the empty SimpleNamespace object is still truthy, so the if doesn't trip! I'll merge after a quick fix.]

PR #38, opened 2022-03-25
Remove unneeded clang unwrapping

jun-sheaf:
As of Bazel 5.0.0, unwrapping is not required for clang on MacOS.

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -176,14 +176,6 @@ def _get_apple_DEVELOPER_DIR():
     # Unless xcode-select has been invoked (like for a beta) we'd expect '/Applications/Xcode.app/Contents/Developer' from xcode-select -p
     # Traditionally stored in DEVELOPER_DIR environment variable, but not provided.
 
-
-@functools.lru_cache(maxsize=None)
-def _get_apple_active_clang():
-    """Get path to xcode-select'd clang version."""
-    return subprocess.check_output(('xcrun', '--find', 'clang'), encoding='utf-8').rstrip()
-    # Unless xcode-select has been invoked (like for a beta) we'd expect '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang' from xcrun -f clang
-
-
 def _apple_platform_patch(compile_args: typing.List[str]):
     """De-Bazel the command into something clangd can parse.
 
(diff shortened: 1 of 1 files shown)


cpsauer:
Hi, @jun-sheaf! Thanks for tossing up a PR and for working with the tool more generally.
I love that you're looking to help us aggressively keep the tool simple and cull unneeded code.

Unless I'm missing something, though, it seems like the wrapping is still needed. I just ran a quick test with Bazel 5.1 on a repo submitted by another user for an issue. And without unwrapping, header finding fails (as I'd have expected), with `Error: DEVELOPER_DIR not set.` (This is with Xcode--are you seeing differently with command line tools?)

Cheers and thanks!
Chris

jun-sheaf:
Ah, this must be because you don't set it independently in your .bashrc (or whatever rc file you use). I suggest that we ask users to set this instead of using this hack since this doesn't work at all with any custom compiling: (solution somewhere in the issue below)

https://github.com/bazelbuild/bazel/issues/10533

We can also set the env variables within the script.

cpsauer:
Hey, Randolf! Thanks for replying back so fast.

I'd really like to understand how this is breaking your custom compiling! Could you tell me more about that? Would love to work together to fix whatever's wrong there. 

Some musings: 
- I know we have lots of users using custom toolchains, so this must be a different, interesting case. 
- I just can't land your code until it won't break the main macOS use case. When possible, I think we want to automatically work well with users' Bazel configurations as they are. 
- To trigger that case, your custom compiling must have the `__BAZEL_XCODE_` flags needing expansion for clangd to understand them, right? Is the issue that it's replacing the compiler? Could we solve your case by adding a condition on the compiler path that only applies the current replacement if it looks like the problematic wrapper script?

Sounds like you've explored this a fair bit. And I'm definitely open to improvements. But here's what I'm worried about with each of the eviironment-based solutions above:
- I would guess the Xcode-based toolchains, as opposed to the cpp-only ones, are important for building for Apple platforms, like macOS or iOS. Is that right? (We need to support, e.g., iOS development.)
- While we can definitely set the wrapper-required environment variables in header finding, we'd also need to take care that clangd will pass them when invoking the wrapper to asking for the default-included headers. Unfortunately, I don't think we can currently embed the variables in the compile_commands.json file (for more, see https://github.com/clangd/clangd/issues/654). And we'd need to take additional care if we wanted to make sure they were always set to the right (differing?) values when the clangd server was launched.

Thanks!
Chris

jun-sheaf:
I've found the issue. I'll submit a new PR for it.

Issue #35, opened 2022-03-24
[Windows] Support --features=compiler_param_file


LoSealL:
Build using `--features=compiler_param_file` would wrap all source files and flags into a `.params` file, I think we should support to read from the `.params` file instead of reading from command line directly.

cpsauer:
Hello again, @LoSealL! Good to see you back :)

I'd love to understand a little more about what you're trying to solve with this. I think [that's a Windows-specific bazel flag](https://docs.bazel.build/versions/main/windows.html) for command lines that are too long. Is that right?

[In my head, I'm imagining that we probably wouldn't need to work around command-line-length limits for clangd, right, since presumably clangd can read arbitrary length commands out of the compile_commands.json file--and we'd like everything self contained in that file?]

All the best,
Chris

LoSealL:
Yes, this is the Windows-specific flag in Bazel, while it's also a recommended flag in Bazel's [official document](https://bazel.build/docs/windows). So I think many people using Bazel on Windows would specify this flag (including me).

It doesn't affect clangd itself, but while parsing sources and headers in the python script, it will fail to read the true parameters because it only gets a parameter file, such as:
```
cl.exe --language=c++ @bazel-out/x64_windows-opt/bin/foo.obj.params"
```

So we need a patch to handle it, basicly to read the actual parameters in the file for later parsing.

**Thanks.**

cpsauer:
Oh I see. (Same link, right?) Is the issue is just for header finding, then?

I'm mostly Linux/mac focused at the moment, as you probably know. So I'll need a bit of your help here.

Do you know if the max command length also exists for subprocess calls from Python?

I'm trying to scope whether it would it work to have this tool just (automatically) turn off that feature--or whether it'll be important to keep it on.

cpsauer:
Windows compatibility hopefully landing soon. It's a bit stalled out because some things need to be a bit more robust before it lands IMO.

LoSealL:
No, all the sources and headers will be lost.
Currently I'm tracking the Windows PR #28 and add all my supports there (maybe through my own PR after #28 was merged)

cpsauer:
I really appreciate your care and willingness to help @LoSealL. 

We'll definitely need to land #28 and then return to this. (I'm going to try to provision a Windows machine tomorrow.)

[...]

PR #49, opened 2022-04-26
support compiler_param_file features on Windows

LoSealL:
Highlights:
- When enable `--feature=compiler_param_file` on command line or **.bazelrc file**, `_get_file` can read the actual compile args in the file
- Fully compatible without this feature or change this feature during the developmemt.

Lowlights:
- When using compile commands extractor along with `--feature=compiler_param_file`, user must build the target once, because `bazel aquery` won't actually generate the `.params` file. (Which I think is a bazel limitation)
- Strip tail whitespaces and fix a typo whose charactor is not encoded as unicode..


cpsauer:
For context for any other readers, this fixes #35

refresh.template.py
```diff
@@ -147,6 +148,17 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
         # End:   template filled by Bazel
     ))
 
+    # Write header_cmd to a temporary file, so we can use it as a parameter to cl.exe,
+    # because Windows cmd has a limitation of 8KB of command line length. So here we make a threshold of len(compile_args) < 8000
+    WIN_CMD_LIMITS = 8000
```
cpsauer, line 153: Worth documenting the 8191--and why you chose a smaller limit? I assume because length calculate is an approximation of the real one with escaping https://docs.python.org/3/library/subprocess.html#converting-an-argument-sequence-to-a-string-on-windows

Another, (maybe better) approach could be try-except whatever exception it throws when the command is too long.

refresh.template.py
```diff
@@ -227,6 +239,28 @@ def _get_headers(compile_args: typing.List[str], source_path: str):
 _get_headers.has_logged_missing_file_error = False
 
 
+def _params_file_to_source_files(compile_args: typing.List[str]):
```
cpsauer, line 242: This and the changes to get_files are no longer needed, right?

refresh.template.py
```diff
@@ -147,6 +148,17 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
         # End:   template filled by Bazel
     ))
 
+    # Write header_cmd to a temporary file, so we can use it as a parameter to cl.exe,
+    # because Windows cmd has a limitation of 8KB of command line length. So here we make a threshold of len(compile_args) < 8000
+    WIN_CMD_LIMITS = 8000
+    if len('\n'.join(header_cmd)) >= WIN_CMD_LIMITS:
+        fd, temp_params = tempfile.mkstemp(text=True)
```
cpsauer, line 155: Per your question in the issue: Let's clean up the temp file when we're done, because there could be a lot of these! How about using one of the other helpers that auto deletes w/ a context manager, like tempfile.NamedTemporaryFile?

refresh.template.py
```diff
@@ -147,6 +148,17 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
...
+        with open(temp_params, 'w') as f:
+            # should skip cl.exe the 1st line
+            f.write('\n'.join(header_cmd[1:]))
```
cpsauer, line 158: Unsure: Do we need a newline at the end?

refresh.template.py
```diff
@@ -147,6 +148,17 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
...
+        os.close(fd)
+        header_cmd = [header_cmd[0], f'@{temp_params}']
```
cpsauer, line 160: Love format strings :)

refresh.template.py
```diff
@@ -414,6 +454,8 @@ def _get_commands(target: str, flags: str):
         # Shush logging. Just for readability.
         '--ui_event_filters=-info',
         '--noshow_progress',
+        # Disable param file, only useful for Windows. No effect on Linux.
```
cpsauer, line 457: Could we document here why we're doing this? eg Not needed for aquery because command is not run, and that we want the command even if the code doesn't build

cpsauer (commented):
Thanks so much for doing, @LoSealL!

I'm delighted that this route is working and really appreciate your good efforts here.
I'd love it if we cleaned things up just a bit before merging, though. Comments inline

refresh.template.py
```diff
@@ -147,6 +148,17 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
         # End:   template filled by Bazel
     ))
 
+    # Write header_cmd to a temporary file, so we can use it as a parameter to cl.exe,
```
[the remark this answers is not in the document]
LoSealL: Yes, I set limit to 1 to force all headers go into this way, and it works for me.

refresh.template.py
LoSealL: Sure, will add better doc here

refresh.template.py
LoSealL: Sure, will remove

refresh.template.py
LoSealL: I intend to do this, but just noticed Bazel leave quite a lot temp files there :-(
Will fix it.

refresh.template.py
LoSealL: No need for cl.exe to work, but a newline at end is better to print and read.

refresh.template.py
LoSealL: Sure

cpsauer:
Thanks so much, @LoSealL!

refresh.template.py
cpsauer: Bazel's are caches, though, hopefully, right? If they're just accumulating piles of temp files on Windows, it might be worth filing a quick issue with them.

refresh.template.py
cpsauer: S.g.  Thoughts on the pro/con of the try-except approach?

LoSealL: try-except will be conflict with the following logic:
```python
for line in header_search_process.stderr.splitlines():
```
We can't detect whether the error comes from compile failure or command line overflow. Not sure if errorcode can help.

cpsauer: Maybe I'm missing something, but I was thinking that there's probably a (Python) Error thrown if the line is longer than the command length limit (after escaping and all that). Like this one https://stackoverflow.com/questions/2381241/what-is-the-subprocess-popen-max-length-of-the-args-parameter?  You could test by exceeding the limit to subprocess.run and seeing what it throws!
(Let's definitely not rely on parsing an error message out of stderr--your hardcoded 8000 is more reliable than that!)

cpsauer: Thanks for doing these, and experimenting around!

LoSealL: @cpsauer Tested on my Windows 10 21H2:
For `powershell`, it can support user to type more than 8191 charactors on the console, while for `cmd`, user can't type any more if command line reaches 8191 charactors.
If you execute command from a script (i.e. `.ps1` or `.cmd/.bat`), the cmd may ignore charactors that exceeds 8191 limitation, while powershell doesn't. For example:

**CMD:**
```batch
set a=1234...(more than 8191 chars)....EOL
:: will only show 1234....(to 8191 length), can't echo "EOL", and no error
echo %a%
```

**PWSH:**
```powershell
$env:a="1234....(more than 8191 chars)...EOL"
# show full $a string, everything is OK
echo $env:a
```

But if your file path is too long (by default it can't longer than 254), most applications will fail with an error (just as your link) with both cmd and powershell console.

cpsauer: Ah, sure, but I think we're talking past each other somehow.

I meant that *Python* would probably throw an Error if the args to subprocess.run were too long. That is, is an error like the one linked. Is such an error thrown if you temporarily replace `header_cmd` in the subprocess command with `"a "*10000` (or similar)?

refresh.template.py
```diff
@@ -147,6 +148,20 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
         # End:   template filled by Bazel
     ))
 
+    # Write header_cmd to a temporary file, so we can use it as a parameter file to cl.exe,
+    # because Windows cmd has a limitation of 8191 charactors. So here we make a threshold of len(compile_args) < 8000
+    # See https://docs.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation
+    WIN_CMD_LIMITS = 8000
+    tempdir = None
+    if len('\n'.join(header_cmd)) >= WIN_CMD_LIMITS:
+        tempdir = tempfile.TemporaryDirectory()
```
cpsauer, line 157: Can I ask why we're doing a TemporaryDirectory instead of, e.g., a context manager over  tempfile.NamedTemporaryFile? Like for example:
`with tempfile.NamedTemporaryFile(...`

(Offhand, I think that'd be simpler than creating and manually destroying a directory, too, but maybe I'm missing something.)

LoSealL: Oh, got your point.
Just test and `subprocess.call` will return error code **1**, with an stderr output "_command line too long_".

We may check return code 1 or stderr text (the text is in locale language)

refresh.template.py
LoSealL: Oh, got your point.
Just test and `subprocess.call` will return error code **1**, with an stderr output "_command line too long_".

We may check return code **1** or stderr text (the text is in locale language)

refresh.template.py
LoSealL: Will change to NamedTemporaryFile

refresh.template.py
LoSealL: Refer to https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/1bc92ddf-b79e-413c-bbaa-99a5281a6c90
msvc compiler (only tested cl.exe) normally returns error code 2 (such as command error, compile error...)
We will use error code 1 for **command line too long** problem

cpsauer: Oh, bummer. So no Python error, then. Sorry for pushing in the wrong direction.

Do whatever you think is safest and most robust here. If you're confident code 1 is always this command-length-issue, then do that. And if not, then let's fall back to your old solution. (Again, sorry.)

refresh.template.py
```diff
@@ -147,14 +148,30 @@ def _get_headers_msvc(compile_args: typing.List[str], source_path: str):
...
+        # To overcome this issue, we can call command parameters from a params file and use the '@' switch to pass the file name.
+        # E.g. cl.exe @params_file.txt
+        # If the return code is 1, it means the command line is too long (>=8191), so we write the command to a temp file and call from it.
+        temp_params = tempfile.NamedTemporaryFile('w')
+        # should skip cl.exe the 1st line
+        temp_params.write('\n'.join(header_cmd[1:]))
+        header_cmd = [header_cmd[0], f'@{temp_params}']
+        header_search_process = _search_headers(header_cmd)
+        # close and delete the temp file we created
+        temp_params.close()
```
cpsauer, line 174: Thanks for documenting things so well, and sharing code via the local fn!

A few last things:
1) I didn't read the docs carefully enough on NamedTemporaryFile the first time. I'm now seeing 
>Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).


...which seems like it would cause problems. How do you think we should solve this? (Perhaps we do need to use tempfile.mkstemp after all...sorry.)
2) I think we want to be careful to make sure that we're flushing the write offer before calling subprocess.

refresh.template.py
```diff
@@ -241,6 +258,8 @@ def _get_files(compile_args: typing.List[str]):
                 # If we did ever go this route, you can join the output from aquery --output=text and --output=jsonproto by actionKey.
             # For more context on options and how this came to be, see https://github.com/hedronvision/bazel-compile-commands-extractor/pull/37
     compile_only_flag = '/c' if '/c' in compile_args else '-c' # For Windows/msvc support
+    if compile_only_flag not in compile_args:
+        raise ValueError(f"{compile_only_flag} not found in compile_args: {compile_args}")
     source_index = compile_args.index(compile_only_flag) + 1
```
cpsauer, line 263: I think index will already throw a ValueError if not found!

LoSealL: It just throws a KeyError, I think this exception will better let user know witch `compile_args` is missing.

refresh.template.py
LoSealL: The doc in the source code says that NamedTemporaryFile does use mkstemp to create file... I'd like to double check that

LoSealL:
```python
def NamedTemporaryFile(mode='w+b', buffering=-1, encoding=None,
                       newline=None, suffix=None, prefix=None,
                       dir=None, delete=True, *, errors=None):
    """Create and return a temporary file.
    Arguments:
    'prefix', 'suffix', 'dir' -- as for mkstemp.
    'mode' -- the mode argument to io.open (default "w+b").
    'buffering' -- the buffer size argument to io.open (default -1).
    'encoding' -- the encoding argument to io.open (default None)
    'newline' -- the newline argument to io.open (default None)
    'delete' -- whether the file is deleted on close (default True).
    'errors' -- the errors argument to io.open (default None)
    The file is created as mkstemp() would do it.

    Returns an object with a file-like interface; the name of the file
    is accessible as its 'name' attribute.  The file will be automatically
    deleted when it is closed unless the 'delete' argument is set to False.
    """

    prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)

    flags = _bin_openflags

    # Setting O_TEMPORARY in the flags causes the OS to delete
    # the file when it is closed.  This is only supported by Windows.
    if _os.name == 'nt' and delete:
        flags |= _os.O_TEMPORARY

    (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
    try:
        file = _io.open(fd, mode, buffering=buffering,
                        newline=newline, encoding=encoding, errors=errors)

        return _TemporaryFileWrapper(file, name, delete)
    except BaseException:
        _os.unlink(name)
        _os.close(fd)
        raise

def _mkstemp_inner(dir, pre, suf, flags, output_type):
    """Code common to mkstemp, TemporaryFile, and NamedTemporaryFile."""

    names = _get_candidate_names()
    if output_type is bytes:
        names = map(_os.fsencode, names)

    for seq in range(TMP_MAX):
        name = next(names)
        file = _os.path.join(dir, pre + name + suf)
        _sys.audit("tempfile.mkstemp", file)
        try:
            fd = _os.open(file, flags, 0o600)
        except FileExistsError:
            continue    # try again
        except PermissionError:
            # This exception is thrown when a directory with the chosen name
            # already exists on windows.
            if (_os.name == 'nt' and _os.path.isdir(dir) and
                _os.access(dir, _os.W_OK)):
                continue
            else:
                raise
        return (fd, _os.path.abspath(file))

    raise FileExistsError(_errno.EEXIST,
                          "No usable temporary file name found")
```

Pasted the code here. It calls `_mkstemp_inner` which is same as `mkstemp`, and it uses a `_RandomNameSequence` to generate temp file name. `TMP_MAX` is 10000 to guarantee there's no name conflict.


Now the test code:
```python
import os
import tempfile
import unittest
from concurrent.futures.thread import ThreadPoolExecutor


def _create_temp(bin):
    with tempfile.NamedTemporaryFile('w') as f:
        bin.append(f.name)
        f.write('abc')


class TestNamedTemporaryFile(unittest.TestCase):
    def test_mt_create(self):
        TEMP_BIN = []
        Q = 1000000

        pool = ThreadPoolExecutor(os.cpu_count())
        [pool.submit(_create_temp, TEMP_BIN) for _ in range(Q)]
        pool.shutdown(wait=True)
        self.assertEqual(len(TEMP_BIN), Q)
        self.assertEqual(len(set(TEMP_BIN)), Q)


if __name__ == '__main__':
    unittest.main()
```
Which is passed here.

cpsauer: Huh, interesting!

 I wonder what part prompted them to add that reopen warning to the docs.

refresh.template.py
cpsauer: [At least on my system]
```Python
>>>"".index("foo")
  File "<stdin>", line 1, in <module>
ValueError: substring not found
```

But regardless, this is going to lead to a hard crash, right? Like this isn't something we expect to show the user.

LoSealL: But w/o this two lines it still goes to a hard crash if both `/c` or `-c` not found...

cpsauer: Totally agree!
It'll always say "-c not found...", though, bc of the previous line, though, right?
Not super opposed or anything, I'm not sure how much this adds.

LoSealL: You're right! What about change typo to 
```python
if compile_only_flag not in compile_args:
  raise ValueError(f"/c or -c (required for parsing sources) is not found in compile_args: {compile_args}")
```

cpsauer: Sure, but I really think this is okay without! They should always be present from Bazel, and we'd get a crash that user will report

LoSealL: OK, I will just delete them. Maybe I only encounter during my development.

cpsauer: Oh! I'm silly. You're adding it so that the contents of the array itself will be in the error message for easy debugging, right? That's awesome, because then people will actually report the compile command as part of the error.

(Sorry I was slow to understand.)

refresh.template.py
cpsauer: Aha! Looks like the thing that causes the problem with NamedTemporaryFile vs mkstemp in the source above is is O_TEMPORARY.

Explained over at  https://stackoverflow.com/questions/15169101/how-to-create-a-temporary-file-that-can-be-read-by-a-subprocess

And flushing comes up there, too.

---

I also want to make sure we aren't accidentally waiting on each other. I'm still waiting on you fixing this one before merging, right? And you're not waiting on me?

refresh.template.py
cpsauer: ^ To be clear, if easy, could you re-add your code from https://github.com/hedronvision/bazel-compile-commands-extractor/pull/49#discussion_r869977295. Again, my apologies for being slow to understand what you were up to.

refresh.template.py
LoSealL: @cpsauer Ah, I got what you mean here, and there's indeed a potentail risk for `cl.exe` to read an opened temp file (while it works for me though...). Let's delete it manually as this post did.

LoSealL:
@cpsauer I updated the commit

cpsauer:
Sweet. But perhaps we should take one of the more robust answers from that stack overflow post?
[Also, any chance we could add a more direct link documenting the error code? I saw the 500 page pdf, but couldn't quickly find it therein.]

cpsauer:
You may have a deeper understanding of what's going on in windows and O_TEMPORARY? Perhaps it's opened safely by msvc anyway? Not sure if you saw the part of that post about being able to reopen it using O_TEMPORARY? Still do think we should make sure we flush, though.

LoSealL:
For `O_TEMPORARY` flag, you can refer to this [page](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-170).
Briefly, trying to open a file created with flag `O_TEMPORARY` and not close yet, you must open with `O_TEMPORARY` as well, or you don't have the permission (**PermissionError**).
On the other hand, open a normal file is OK for both read and write, because the default flag is `O_RDWR`.

LoSealL:
What do you mean by referring **more robust answer**? Can you paste the code?

And for CL error code, as this [page](https://docs.microsoft.com/en-us/cpp/build/reference/return-value-of-cl-exe?view=msvc-170) said, it is the error code of Windows and there are lots of possible values...

cpsauer:
Hey @LoSealL: I finally got back to my Windows machine and took a shot at the remaining things.

I think we definitely are all set on turning off compiler_param_file to aquery. ☑️ 

For spilling arguments to command files when things get too long:
At least on my machine, I was seeing checking the return value fail because subprocess.run was instead throwing winerror 206 when the command got too long.  So I redid things around catching that, and made the cleanup a bit more exception tolerant, etc. Interestingly, I was definitely seeing errors only for commands well over 8k. More like 30k.

That said, I'm still far from confident that we're handling all the cases right. For example: Do we need to be quoting the arguments in the param file? Should they maybe be all on one line to avoid splitting arguments that have connected values? If your logic was working on your machine, why did I need the try-except on mine? Is my .encode() [to utf-8, I believe by default] going to cause problems with your different code page?

Thoughts?


---

Also, separate, but any chance I could ask you to give the new `exclude_headers = "all"`, `exclude_headers = "external"`, `exclude_external_sources = True` a test on Windows, just to make sure they're working well there?

Thanks so much,
Chris

LoSealL:
Hi @cpsauer . You are right.

Following your finds, I create a sample repo [here](https://github.com/LoSealL/bazel-command-extractor-sample-windows). Which shows different behaviour depends on executables.
And for `CL.exe`, I think your solution is OK.

You can check that repo first to see if we still have diverges.

LoSealL:
@cpsauer Updated:

I even can't reproduce my issue today, which I thought could caused by compile and link too many source files. Maybe it's solved by upgrading VS version?

So for now, we can make a conclusion here:
1. For compiler flags too long:
        * users **need** `--features=compiler_param_file`, and without it, bazel will report _"command is longer than CreateProcessW's limit (32767 characters)"_ .
        * If user call `cl.exe` directly from command line, it worked somehow.
        * If user call `cl.exe` from python subprocess module, it will throw an exception with Windows error code 206.

2. For linker flags too long or too many files to be compiled:
        * Both should work normally in bazel build. No need any features.

LoSealL:
> * Do we need to be quoting the arguments in the param file? ([Bazel seems to do this, but always uses gcc format](https://github.com/bazelbuild/bazel/blob/a48e246e53c19cfa5d68c21ffefee0470acf7230/src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java#L1436)(!))

Add more tests to [sample repo](https://github.com/LoSealL/bazel-command-extractor-sample-windows), please check there.

Bazel itself can't support path with space with `compiler_param_file` feature, so my conclusion is **no need**. If you find more escape cases, please add there to help us to check.

> * Should they maybe be all on one line to avoid splitting arguments that have connected values? (Bazel seems to separate with newlines.)

Currently no issues found here.

> * If your logic was working on your machine, why did I see the winerror at a higher limit and need the try-except on mine? (If this is thorny, one resolution would be to _always_ spill over to a param file.)

Sorry, I can't reproduce either today. So it's right to stick to your solution.

> * Is my .encode() [to utf-8, I believe by default] going to cause problems with your different code page? I assume so, so we'd better switch to os.fdopen

For temp file that write and read internally, utf-8 is OK, because we both write and read with utf-8. (CL.exe reads utf-8 normally).

Issues only happen when python reads a non-utf file. (i.e. a native notepad saved file, or echo command flush to a file with non utf-8 codepage.)

> * etc.

cpsauer:
Great! Thanks for working to clarify things, @LoSealL.

I quickly added some escaping in case Bazel ever fixes its implementation. And then I merged this in.

**Could I ask you to proofread the latest version, test it, and report back?**
I'd really appreciate it if you'd trigger the param file generation to make sure it all works. And could you also check out the "exclude" features above? (I'm back away from my Windows dev machine, unfortunately.)

LoSealL:
Sure, also my many thanks to your great support!

cpsauer:
And thank you for sticking with it! I know it's been a long adventure, and I really appreciate your help, start to finish

cpsauer:
Some last miscellaneous things: 

> bazel will report "command is longer than CreateProcessW's limit (32767 characters)"

Just to confirm: You can aquery successfully without the `compiler_param_file` feature, right?  And this only happens when running build actions?

And in your test repo, I'm wondering if the following explains what was going on:
`echo` -> Special because it's a builtin, so it envokes cmd.exe -> 8k command length limit and special message
For the others: 32k limit. Only cl.exe was over the limit because the parameters were shorter for the others?

cpsauer:
Anyway, please do tell me if you have any concerns with what we've got here! I'd love to hear them.

And mostly, thank you for all your efforts and for checking and testing!

cpsauer:
@LoSealL, could I ask one more Windows favor?
Could you run a bazel build on whatever repo you're using and see if there are files listing headers and other dependencies next to the .obj files? (under bazel-out) And if so, could you send me one, along with its path?

[To explain the importance: I'm trying to speed up the header finding by examining dependency files Bazel has cached. On macOS/Linux, there are .d files, listing headers in makefile format, and I'd like to know if there's something similar on Windows. https://github.com/hedronvision/bazel-compile-commands-extractor/commit/fe5e048c55a848b0b3652b8975463a13454e9e79 is the commit if you're curious.]

LoSealL:
@cpsauer No, there's no such `.d` like file on Windows. Instead, a `.param` file if `--features=compiler_param_file`, and the param file is a compiler command flags.

PR #58, opened 2022-06-17
Allow optional space between target and colon

abcdabcd987:
I triggered the following assertion when generating the compile commands for TensorFlow.

```
  File "/.../execroot/org_tensorflow/bazel-out/k8-opt/bin/refresh_compile_commands.runfiles/org_tensorflow/refresh_compile_commands.py", line 90, in _parse_headers_from_makefile_deps
    assert split[0].endswith('.o:'), "Something went wrong in makefile parsing to get headers. Zeroth entry should be the object file. Output:\n" + d_file_content
AssertionError: Something went wrong in makefile parsing to get headers. Zeroth entry should be the object file. Output:
list_kernels.cu.o : tensorflow/core/kernels/list_kernels.cu.cc \
    /usr/include/stdc-predef.h \
    /usr/local/cuda-11.2/bin/../targets/x86_64-linux/include/cuda_runtime.h \
   (... omit hundreds of lines ...)
```

It seems that the makefile could take both formats: `target: deps` and `target : deps`, i.e., there can be optional spaces between the target and the colon. This pull request fixes the issue by split out the target explicitly.

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -83,8 +83,11 @@ def _parse_headers_from_makefile_deps(d_file_content: str, source_path_for_sanit
 
     See https://clang.llvm.org/docs/ClangCommandLineReference.html#dependency-file-generation for more.
     """
-    split = d_file_content.replace('\\\n', '').split() # Undo shell line wrapping bc it's not consistent (depends on file name length). Also, makefiles don't seem to really support escaping spaces, so we'll punt that case https://stackoverflow.com/questions/30687828/how-to-escape-spaces-inside-a-makefile
-    assert split[0].endswith('.o:'), "Something went wrong in makefile parsing to get headers. Zeroth entry should be the object file. Output:\n" + d_file_content
+    colon_idx = d_file_content.index(':')
+    target = d_file_content[:colon_idx].strip()
+    split = [target]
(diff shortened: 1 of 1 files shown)


cpsauer:
Hey @abcdabcd987! 

Thanks for playing with this tool and for taking the time to leave it better than you found it. Looks great, and I really appreciate your broadening the usefulness of the tool.

Probably our first use of cuda! I assume this makefile was probably outputted by NVCC, and that nvcc therefore must be deciding to put a space before the colon.

(I'll merge in one sec--and then might massage just a little bit--so if I've broken things somehow, just holler.)

Again, thanks so much for contributing!

Cheers,
Chris

cpsauer:
Looking at the NVCC docs, I'm guessing instead that your underlying command is compiling cuda via clang, since cuda lists support for a different set of arguments. But if that's wrong, and you are using cuda, could you check for me that the hundreds of lines, omitted above, are a list of headers in makefile format and *not* e.g. the preprocessed code being dumped to stdout?

cpsauer:
@abcdabcd987, just swinging back around to check in. could I ask you to confirm that the latest commit is working for you and, if easy, answer some of the questions above?

Issue #30, opened 2022-02-23
my repo already has external directory...


alkis:
but that does not point to external `bazel-out/.../../../external`. It would be nice if the name of this link could be configurable in `refresh_compile_commands`. Alternatively the script could be smarter to look in several possible paths for external: `external`, `bazel-external`, `vendor`, `bazel-vendor`.

alkis:
I got this to work... Here's the diff:

```
diff --git a/refresh.template.py b/refresh.template.py
index 6486fab..8ee52db 100644
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -221,7 +221,8 @@ def _all_platform_patch(compile_args: typing.List[str]):
     # For more context see: https://github.com/hedronvision/bazel-compile-commands-extractor/issues/21
     compile_args = (arg for arg in compile_args if not arg == '-fno-canonical-system-headers')
 
-    # Any other general fixes would go here...
+    ext = _all_platform_patch.external
+    compile_args = (ext + arg[8:] if arg.startswith('external/') else arg for arg in compile_args)
 
     return list(compile_args)
 
@@ -348,6 +349,8 @@ if __name__ == "__main__":
         # End: Command template filled by Bazel
     ]
     compile_command_entries = []
+    _all_platform_patch.external = "@EXTERNAL@"
+
     for (target, flags) in target_flag_pairs:
         compile_command_entries.extend(_get_commands(target, flags))
 
diff --git a/refresh_compile_commands.bzl b/refresh_compile_commands.bzl
index 5514c9d..a4d553b 100644
--- a/refresh_compile_commands.bzl
+++ b/refresh_compile_commands.bzl
@@ -36,7 +36,7 @@ refresh_compile_commands(
 ########################################
 # Implementation
 
-def refresh_compile_commands(name, targets = None):
+def refresh_compile_commands(name, targets = None, external = None):
     # Convert the various, acceptable target shorthands into the dictionary format
     if not targets: # Default to all targets in main workspace
         targets = {"@//...": ""}
@@ -47,7 +47,7 @@ def refresh_compile_commands(name, targets = None):
 
     # Generate runnable python script from template
     script_name = name + ".py"
-    _expand_template(name = script_name, labels_to_flags = targets)
+    _expand_template(name = script_name, labels_to_flags = targets, external = external)
     native.py_binary(name = name, srcs = [script_name])
 
 
@@ -58,13 +58,17 @@ def _expand_template_impl(ctx):
[...]

Issue #14, opened 2021-12-28
bug: `extract.py` can't deal with non-standard `--symlink_prefix`s


lummax:
When setting [`--symlink_prefix`] on the commandline or `.bazelrc`, the `bazel-out` symlink might not exist or be somewhere else. But `extract.py` currently assumes `bazel-out` exists and generates a `compile_commands.json` file containing many references to it.

There seems to be no way of accessing the `symlink_prefix` via an environment variable or similar, so the `refresh_compile_commands()` might need an additional argument.

[`--symlink_prefix`]: https://docs.bazel.build/versions/main/command-line-reference.html#flag--symlink_prefix

cpsauer:
Hi, Lukas! Thanks for giving the tool a try and pointing out an interesting case I hadn't considered. Sorry things aren't working out of the box. Let's figure it out :)

I've been trying to reproduce this one over here--and the behavior I'm seeing is that even with `--symlink_prefix=whatever` or `--symlink_prefix=`, `bazel-out` still gets created on run, keeping the paths valid. (Which would check out against, e.g. https://github.com/bazelbuild/bazel/issues/3796. I'm doing, for example, `bazel run :refresh_all --symlink_prefix=foo` on both bazel latest and bazel rolling). You're for sure seeing otherwise? If so, anything else I should know or try?

If it does persists, I think this is the easiest way to get things working:
Given what you say about not being able to access `symlink_prefix`, this one seems like it might be better worked around, rather than doing duplicate configuration in the rule.
What about just manually creating the bazel-out symlink to point through your custom named symlink, similar to what we did with `external` in the README? (I presume you can't remove the flag and that it serves some essential function. Would love to learn what you're using it for!) 

Thanks!
Chris

P.S. I see Bremen as your location. I've been there, and have some family relatively nearby :)

lummax:
Hey, sorry you are right, my setup above was incomplete. We actually have `build --symlink_prefix=custom/prefix --experimental_no_product_name_out_symlink` in our `.bazelrc` (running on `bazel` 4.2.1).

We are in a process of migrating a code-base using `make` and `gprbuild`/`msbuild` to `bazel`. There are some 
[...]

PR #65, opened 2022-07-07
Add options `replace_external_path` / `replace_output_path` to no lon…

lummax:
…ger require symlinks

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -10,6 +10,7 @@ Interface (after template expansion):
     - Crucially, this output is de-Bazeled; The result is a command that could be run from the workspace root directly, with no Bazel-specific requirements, environment variables, etc.
 """
 
+import platform
 import sys
 if sys.version_info < (3,7):
     sys.exit("\n\033[31mFATAL ERROR:\033[0m Python 3.7 or later is required. Please update!")
@@ -588,8 +589,23 @@ def _all_platform_patch(compile_args: typing.List[str]):
 
     return list(compile_args)
 
+def _apply_path_replacements(compile_args: typing.List[str], replacements: typing.Mapping[str, str]) -> typing.List[str]:
+    def replace_single(arg):
+        for (prefix, replacement) in replacements.items():
+            if arg.startswith(prefix):
+                return replacement + arg[len(prefix) :]
+            for variant in ("={}", "/I{}"):
+                if variant.format(prefix) in arg:
+                    return arg.replace(variant.format(prefix), variant.format(replacement))
+        return arg
 
-def _get_cpp_command_for_files(compile_action):
+    res = (replace_single(arg) for arg in compile_args)
+    if platform.system() == "Windows":
+        res = (arg[0] + arg[1:].replace("/", "\\") for arg in res)
+    return list(res)
+
+
+def _get_cpp_command_for_files(compile_action, replacements: typing.Mapping[str, str]):
     """Reformat compile_action into a compile command clangd can understand.
 
     Undo Bazel-isms and figures out which files clangd should apply the command to.
@@ -598,13 +614,14 @@ def _get_cpp_command_for_files(compile_action):
     compile_action.arguments = _all_platform_patch(compile_action.arguments)
     compile_action.arguments = _apple_platform_patch(compile_action.arguments)
     # Android and Linux and grailbio LLVM toolchains: Fine as is; no special patching needed.
+    compile_action.arguments = _apply_path_replacements(compile_action.arguments, replacements)
 
     source_files, header_files = _get_files(compile_action)
 
     return source_files, header_files, compile_action.arguments
 
 
-def _convert_compile_commands(aquery_output):
+def _convert_compile_commands(aquery_output, replacements: typing.Mapping[str, str]):
     """Converts from Bazel's aquery format to de-Bazeled compile_commands.json entries.
 
     Input: jsonproto output from aquery, pre-filtered to (Objective-)C(++) compile actions for a given build.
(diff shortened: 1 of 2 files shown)


cpsauer:
Hey, Lukas! Good to hear from you. Thanks for your PR--and all the ways you've helped make this project better.

I know we'd discussed this some, way back in #14. Lots of the other things we discussed there have now happened, thanks to the good efforts of you and others--including a robust set of features around header extraction and external source files.

Focusing on some high level things for tonight.
[It got quite late my time, unfortunately, by the time I was able to get to this.] 
- So I understand: What motivates people to need to turn off those symlinks?
  - [I'm not against it. We just haven't needed it, and I want to understand the root need that's motivating the additional implementation complexity and interface.]
- I think getting this robust is going to be a bit subtle; it's harder getting things to work when the workspace root doesn't mirror the compilation environment.
  - As an example: Looking quickly back at 14, there's the `-Ibazel-out/<path>` case. And glancing at a compile_commands.json in front of me, I see an `-isystemexternal/<...>`. I'm sure there are more of these cases, including on Windows, and I'd need you to get them all.
  - We'll also need to take care to not break other key features, like, `_file_is_in_main_workspace_and_not_external` and hitting bazel's .d dependency cache under clang/gcc. There's also the case of file paths being absolute on Windows, and with varying separators.
  - That isn't an exhaustive list, I'm sure; we'd be changing a pretty fundamental, simplifying assumption here, and that's going to need some additional care. Writing this out made me realize that this is even trickier than I'd thought.
 
^ Given the above, I think we'll want to be pretty convinced this is needed and robust before we'd merge. I am probably going to need to ask you to do most of that work and maintainance. Is that worth it? Offhand, I'm tempted to try to talk you into just having the symlinks, especially if the use case is the temporary migration you'd mentioned in #14, but I could be missing something. Hopefully that's not too frustrating; I really appreciate your efforts and am looking forward to your reply.

Cheers,
Chris

--- 
[We shouldn't let these distract from the high level discussion, but here are notes about some more minor things that occurred along the way, so I don't forget them.]

- Maybe we shouldn't gitignore entries when their addition is disabled by a flag? e.g. external and bazel-*.
- Thoughts on leaning toward automatically configuring, like noted in 14? For example, we could detect if bazel-out exists -- and automatically apply that replacement if it does not? 
- Maybe let's consistently detect windows one way or another. Doesn't matter to me whether that's platform.system or os.name. (Also, minor, but import position.) Similar consistency note with _call_bazel.
- If we change the assumptions here, we should take a crack at updating the comments that rely on them. That'll also help find the cases we've broken.

lummax:
Hey Christopher, 

I was delighted to see all the improvements now that I got some time to work on the `compile_commands.json` generation for us. Great work :+1: 

> So I understand: What motivates people to need to turn off those symlinks?

It is mainly Windows support and the inability of various tools there to copy with symlinks. Multiple colleges noticed that even the Windows Explorer search feature will follow symlinks and possibly end up in a symlink loop. Same for some indexing/search implementation for IDEs or command line search utilities. So to keep the Windows people happy we got rid of all bazel symlinks.

> I think getting this robust is going to be a bit subtle; [...] And glancing at a compile_commands.json in front of me, I see an -isystemexternal/<...>. I'm sure there are more of these cases, including on Windows, and I'd need you to get them all.

Yes, I am afraid you are right here. I'm not sure we can find them all, especially for esoteric compilers / configurations. But I'll try my best to find some more cases and add them.

> We'll also need to take care to not break other key features, like, _file_is_in_main_workspace_and_not_external and hitting bazel's .d dependency cache under clang/gcc.

Yes. I would hope that this remains an opt-in feature. We could even name the flags `experimental_...` to make sure users know there are caveats.

>  I am probably going to need to ask you to do most of that work and maintainance. Is that worth it?

Happy to do it, just ping me on related issues.

> Offhand, I'm tempted to try to talk you into just having the symlinks, especially if the use case is the temporary migration you'd mentioned in https://github.com/hedronvision/bazel-compile-commands-extractor/issues/14, but I could be missing something. 

I personally am all in for symlinks. But unfortunately that is not my call because of the mentioned Windows problems. The "temporary" migration will continue for some time. And I guess the Windows problems will remain valid after that.

I'll get back to this PR next week and push some updates.

cpsauer:
> So to keep the Windows people happy we got rid of all bazel symlinks.

Oh, very interesting, this need on Windows. Thank you for explaining what's motivating this! For my learning, could I ask you to expand a little more? 
- Is it that these Windows tools are just treating junctions like normal directories, and therefore searching/copying their contents?
  - [I'd have guessed, perhaps, that people might want to automatically search/index their external and generated code. I'm always a little bummed that VSCode doesn't, e.g., automatically quick open through symlinks on other platforms.]
  - What breaks when copying something containing a junction, if not a deep copy?
- I'd also love to understand how one ends up in a junction loop on Windows with Bazel. (I'm asking only because I've failed to imagine how Bazel would construct one, but I believe you.)

> especially for esoteric compilers / configurations

I think we need only target msvc and gcc/clang flags here. This is making me wonder, if you think the need is Windows-focused--maybe the flag should be, too? Up to you. It's possible there's also a more general way to structure the replacement.

Really, though, I think we'll need this to be very clearly solid to merge in, given the complexity of removing the simplifying assumption that bazel's commands can apply, almost unchanged, to the workspace. I don't think we should merge experimental-and-broken; and I want to make sure I'm being as up front about that as possible. And while I'm concerned about how hard it might be (see also #30), I really appreciate your identifying the need and dedication to chasing it down. Good luck!

Cheers!
Chris

lummax:
> Is it that these Windows tools are just treating junctions like normal directories, and therefore searching/copying their contents? 

I actually did not dig into this.

> I'd also love to understand how one ends up in a junction loop on Windows with Bazel. (I'm asking only because I've failed to imagine how Bazel would construct one, but I believe you.)

Now that you are mentioning it, this might actually be our fault: We have a `build/` directory in the workspace root that contains generated files from our legacy build system. I configured `bazel` do create the convenience symlinks below `build/` to keep the workspace root as uncluttered as possible. But I also had to `export_files()` some files below `build/` as they were generated by the legacy build-system. That resulted in the symlink cycle.

Thank you for that insight :)

In my mind this a use case the commands extractor should not support. This is a mis-use of `bazel` to some extent. I operated under the assumption that everyone will run into this on windows. Closing this PR :)

PR #75, opened 2022-08-27
Start to make things unit-testable

jfirebaugh:
Before getting too serious about https://github.com/hedronvision/bazel-compile-commands-extractor/issues/14, I want to have at least some confidence that I can make changes to command reformatting without breaking existing functionality. In short, I would like to start to build up a unit test suite for a subset of code that is (mostly) side-effect free. In principle, this is the easiest sort of code to test: you push a command line in, get results back, and assert they are equal to an expected value. To get to that point, we need to be able to mock out side-effecting code like running subprocesses and memoization. I've tried to split up the code accordingly, although this is as yet a very rough cut. Please take a look and let me know if you agree with the general direction. (NB: I'm not fluent in Python so there's probably conventions I'm not following properly.)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -13,4 +13,17 @@ refresh_compile_commands(
 # Implementation:
 # If you are looking into the implementation, start with the overview in ImplementationReadme.md.
 
-exports_files(["refresh.template.py"]) # For implicit use by refresh_compile_commands.
+# For implicit use by refresh_compile_commands.
+exports_files([
+    "refresh.template.py",
+    "command_reformatter.py",
+])
+
+
+py_test(
+    name = "command_reformatter_test",
+    srcs = [
+        "command_reformatter.py",
+        "command_reformatter_test.py",
+    ],
+)
--- a/command_reformatter.py
+++ b/command_reformatter.py
@@ -1,0 +1,496 @@
+import sys
+import json
+import locale
+import os
+import pathlib
+import re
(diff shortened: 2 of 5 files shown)


cpsauer:
Hey, John! Good to hear from you. Thanks for your good efforts, enthusiasm, and communication.

My first gut reaction is: I'm not sure how much unit testing is going to ease the key problem for #14, which I think is the trickiness of thinking through the new cases the feature brings up. On board with the value of unit tests bring when optimizing/refactoring the implementation of a fixed interface. But regressions on independent sub units hasn't been a meaningful problem here so far, and I'm doubtful it will for this new feature. Almost all the code changes have instead been about a new, rarer case that pops up, with lots of quick, fluid changes to intermediate interfaces. Asserts/warnings have been a super valuable tool so far, since they help check for unanticipated cases across a wide swath of users.

If we do go the unit testing route, my co-founder (who's more a python expert than I) is reminding me that Python is dynamic enough that one can often avoid the dependency inject/split test patterns that you see in, e.g., Java. One can just import, overwrite the functions you want to mock (`imported.foobarfn = <my mock>`). pytest is a good alternative to unittest, he says!

Cheers,
Chris

jfirebaugh:
Okay! I think it's probably best for me to just maintain local patches that makes it work for us and not try to upstream my changes.

PR #78, opened 2022-09-12
Fix "missing from graph" shushing with timestamps

lpulley:
The `WARNING: Targets were missing from graph` shushing doesn't work with Bazel's `--show_timestamps`. This should fix that.


cpsauer:
Hi @lpulley! Thanks for using the tool and contributing. Hadn't thought about this case; I appreciate your handling it!

[I'll briefly add a note about why as a comment and then merge.]

Again thank you! Welcome to the contributors list :)

PR #79, opened 2022-09-13
Fix "missing from graph" shushing with color

lpulley:
The `WARNING: Targets were missing from graph` shushing doesn't work with Bazel's `--color=yes`. This should fix that.


cpsauer:
Thanks for fixing these!

Issue #1, opened 2021-11-17
[Self-Filed] Could we also support Swift?


cpsauer:
**Status:** There's an Apple project to (basically) add Swift support to clangd [here](https://github.com/apple/sourcekit-lsp). It doesn't look mature enough yet (2/2021), but perhaps it will be by the time we'd need it.

**Suggested Workaround:** Use Tulsi for now. 
(Hedron Friendly Faces should use our internal XcodeAdapter wrapper for Tulsi)

cpsauer:
[Intentional blank comment, added to make it clear from the table view that this isn't a user issue lacking a response.]

yrom:
Is there any progress?

cpsauer:
Hey, @yrom! Nope, since you're the first to express interest here. Let me check in on the Apple project.

Looks like the Apple project (linked above) is still listed as early development, but is also now bundled in Xcode and has a fairly popular, highly rated vscode extensions. So maybe a good time to explore. And there's no harm to supporting it early.

I probably shouldn't just build this myself right now only because there's too much else going on. (Need to focus on the startup I'm running, and we don't currently use Swift.) But I'd be delighted to help if you'd be interested in adding this, especially since you just submitted a great pr :)  I think integrating it into this tool would be be the way to go, much better and easier than doing separately.

A high level sketch of what I'd imagine could be a good way of doing it:
1. First, it'd be valuable to check that the Apple project looks to be working well enough to integrate--and needs an integration. I'd probably spin up some quick tests: 
    i. First a non-bazel project using their vscode plugin and making sure that it was supplying decent autocomplete, etc.
    ii. Then using their tool with an existing bazel workspace to check on its compile_commands.json loading, checking that there's a strong need to add compile_commands.json entries to get better autocomplete, and figuring out how to get their indexing during building to work with Bazel.
2. Assuming everything looks good, add the swift compile to our aquery mnemonics filter, separating out those commands so we don't try to pass them to clang to do header extraction, but directing them into compile_commands.json
3. And finally, we'd add some docs so other people can figure out how to use it, too!

-CS

xinzhengzhang:
[...]

PR #89, opened 2022-11-24
Add support for swift

xinzhengzhang:
Add support for `SwiftCompile` in [rule_swift](https://github.com/bazelbuild/rules_swift)

It extracts swift compile commands into `compile_commands.json` providing to [sourcekit-lsp](http://github.com/apple/sourcekit-lsp)

---

VSCode setup
```
code --install-extension sswg.swift-lang
```

Examples:
```BUILD
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
load("@hedron_compile_commands//:refresh_compile_commands.bzl", "refresh_compile_commands")

swift_library(
    name = "Sources",
    srcs = [
        "ViewController.swift",
    ],
    deps = [
         #objc_library
        "//lib:hello",
    ]
)

refresh_compile_commands(
    name = "refresh",
    enable_swift = True,
    targets = {
      "//:Sources": "--apple_platform_type=ios --cpu=ios_arm64",
    }
)
```

```Shell
# Create compile_commands.json
bazel run //:refresh
# Build .swiftmodule
bazel build //:Sources --apple_platform_type=ios --cpu=ios_arm64
# Open is vscode
code .
```

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -917,11 +917,6 @@ def _get_commands(target: str, flags: str):
     Try adding them as flags in your refresh_compile_commands rather than targets.
     In a moment, Bazel will likely fail to parse.""")
 
-    support_mnemonics = ["Objc", "Cpp"]
-    if {enable_swift}:
-        support_mnemonics += ["Swift"]
-    mnemonics_string = '|'.join(support_mnemonics)
-
     # First, query Bazel's C-family compile actions for that configured target
     aquery_args = [
         'bazel',
@@ -929,7 +924,7 @@ def _get_commands(target: str, flags: str):
         # Aquery docs if you need em: https://docs.bazel.build/versions/master/aquery.html
         # Aquery output proto reference: https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/analysis_v2.proto
         # One bummer, not described in the docs, is that aquery filters over *all* actions for a given target, rather than just those that would be run by a build to produce a given output. This mostly isn't a problem, but can sometimes surface extra, unnecessary, misconfigured actions. Chris has emailed the authors to discuss and filed an issue so anyone reading this could track it: https://github.com/bazelbuild/bazel/issues/14156.
-        f"mnemonic('({mnemonics_string})Compile',deps({target}))",
+        f"mnemonic('(Objc|Cpp{swift_mnemonic_string})Compile',deps({target}))",
         # We switched to jsonproto instead of proto because of https://github.com/bazelbuild/bazel/issues/13404. We could change back when fixed--reverting most of the commit that added this line and tweaking the build file to depend on the target in that issue. That said, it's kinda nice to be free of the dependency, unless (OPTIMNOTE) jsonproto becomes a performance bottleneck compated to binary protos.
         '--output=jsonproto',
         # We'll disable artifact output for efficiency, since it's large and we don't use them. Small win timewise, but dramatically less json output from aquery.
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -723,7 +723,7 @@ def _apple_platform_patch(compile_args: typing.List[str], environmentVariables =
         # Bazel wraps the compiler as `external/local_config_cc/wrapped_clang` and exports that wrapped compiler in the proto. However, we need a clang call that clangd can introspect. (See notes in "how clangd uses compile_commands.json" in ImplementationReadme.md for more.)
         # Bazel wrapps the swiftc as `external/build_bazel_rules_swift/tools/worker/worker swiftc ` and worker has been removed in apple_swift_patch
         # Removing the wrapper is also important because Bazel's Xcode (but not CommandLineTools) wrapper crashes if you don't specify particular environment variables (replaced below). We'd need the wrapper to be invokable by clangd's --query-driver if we didn't remove the wrapper.
-        
+
         if compile_args[0].endswith('swiftc'):
             compile_args[0] = 'swiftc'
         else:
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -107,8 +107,8 @@ def _warn_if_file_doesnt_exist(source_file):
         You can either use a refresh_compile_commands rule or the special -- syntax. Please see the README.
         [Supplying flags normally won't work. That just causes this tool to be built with those flags.]
     Continuing gracefully...""")
-        return False
-    return True
+        return True
+    return False
 _warn_if_file_doesnt_exist.has_logged_missing_file_error = False
 
 @functools.lru_cache(maxsize=None)
@@ -594,7 +594,7 @@ def _get_files(compile_action):
     assert source_file.endswith(_get_files.source_extensions), f"Source file candidate, {source_file}, seems to be wrong.\nSelected from {compile_action.arguments}.\nPlease file an issue with this information!"
 
     # Warn gently about missing files
-    if not _warn_if_file_doesnt_exist(source_file):
+    if _warn_if_file_doesnt_exist(source_file):
         return {source_file}, set()
 
     # Note: We need to apply commands to headers and sources.
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -704,14 +704,7 @@ def _apple_swift_patch(compile_args: typing.List[str]):
     # We need to remove it (build_bazel_rules_swift/tools/worker/worker)
     compile_args.pop(0)
 
-    # The worker also expand the arguments defined in swift_rule which was start with -Xwrapped-swift
-    # Expand -debug-prefix-pwd-is-dot
-    match = next((i for i,v in enumerate(compile_args) if v == "-Xwrapped-swift=-debug-prefix-pwd-is-dot"), None)
-    if match:
-        compile_args[match] = "-debug-prefix-map"
-        compile_args.insert(match + 1, os.environ["BUILD_WORKSPACE_DIRECTORY"] + "=.")
-
-    # Remove other -Xwrapped-swift arguments like `-ephemeral-module-cache` `-global-index-store-import-path`
+    # Remove other -Xwrapped-swift arguments like `-debug-prefix-pwd-is-dot` `-ephemeral-module-cache` `-global-index-store-import-path`
     # We could override index-store by config sourcekit-lsp
     compile_args = [arg for arg in compile_args if not arg.startswith('-Xwrapped-swift')]
 
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -93,6 +93,23 @@ def _print_header_finding_warning_once():
     Continuing gracefully...""")
 _print_header_finding_warning_once.has_logged = False
 
+def _warn_if_file_doesnt_exist(source_file):
+    if not os.path.isfile(source_file):
+        if not _warn_if_file_doesnt_exist.has_logged_missing_file_error: # Just log once; subsequent messages wouldn't add anything.
+            _warn_if_file_doesnt_exist.has_logged_missing_file_error = True
+            log_warning(f""">>> A source file you compile doesn't (yet) exist: {source_file}
+    It's probably a generated file, and you haven't yet run a build to generate it.
+    That's OK; your code doesn't even have to compile for this tool to work.
+    If you can, though, you might want to run a build of your code.
+        That way everything is generated, browsable and indexed for autocomplete.
+    However, if you have *already* built your code, and generated the missing file...
+        Please make sure you're supplying this tool with the same flags you use to build.
+        You can either use a refresh_compile_commands rule or the special -- syntax. Please see the README.
+        [Supplying flags normally won't work. That just causes this tool to be built with those flags.]
+    Continuing gracefully...""")
+        return False
+    return True
+_warn_if_file_doesnt_exist.has_logged_missing_file_error = False
 
 @functools.lru_cache(maxsize=None)
 def _get_bazel_cached_action_keys():
@@ -577,19 +594,7 @@ def _get_files(compile_action):
     assert source_file.endswith(_get_files.source_extensions), f"Source file candidate, {source_file}, seems to be wrong.\nSelected from {compile_action.arguments}.\nPlease file an issue with this information!"
 
     # Warn gently about missing files
-    if not os.path.isfile(source_file):
-        if not _get_files.has_logged_missing_file_error: # Just log once; subsequent messages wouldn't add anything.
-            _get_files.has_logged_missing_file_error = True
-            log_warning(f""">>> A source file you compile doesn't (yet) exist: {source_file}
-    It's probably a generated file, and you haven't yet run a build to generate it.
-    That's OK; your code doesn't even have to compile for this tool to work.
-    If you can, though, you might want to run a build of your code.
-        That way everything is generated, browsable and indexed for autocomplete.
-    However, if you have *already* built your code, and generated the missing file...
-        Please make sure you're supplying this tool with the same flags you use to build.
-        You can either use a refresh_compile_commands rule or the special -- syntax. Please see the README.
-        [Supplying flags normally won't work. That just causes this tool to be built with those flags.]
-    Continuing gracefully...""")
+    if not _warn_if_file_doesnt_exist(source_file):
         return {source_file}, set()
 
     # Note: We need to apply commands to headers and sources.
@@ -618,7 +623,6 @@ def _get_files(compile_action):
         compile_action.arguments.insert(1, lang_flag)
 
     return {source_file}, header_files
-_get_files.has_logged_missing_file_error = False
 # Setup extensions and flags for the whole C-language family.
 # Clang has a list: https://github.com/llvm/llvm-project/blob/b9f3b7f89a4cb4cf541b7116d9389c73690f78fa/clang/lib/Driver/Types.cpp#L293
 _get_files.c_source_extensions = ('.c', '.i')
@@ -663,7 +667,7 @@ def _get_apple_SDKROOT(SDK_name: str):
     # Traditionally stored in SDKROOT environment variable, but not provided by Bazel. See https://github.com/bazelbuild/bazel/issues/12852
 
 
-def _get_apple_platform(compile_args: typing.List[str]):
+def _get_apple_platform(compile_args: typing.List[str], environmentVariables = None):
     """Figure out which Apple platform a command is for.
 
     Is the name used by Xcode in the SDK files, not the marketing name.
@@ -674,6 +678,13 @@ def _get_apple_platform(compile_args: typing.List[str]):
         match = re.search('/Platforms/([a-zA-Z]+).platform/Developer/', arg)
         if match:
             return match.group(1)
+    if environmentVariables:
+        match = next(
+            filter(lambda x: x.key == "APPLE_SDK_PLATFORM", environmentVariables),
+            None
+        )
+        if match:
+            return match.value
     return None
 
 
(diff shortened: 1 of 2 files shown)


refresh_compile_commands.bzl
```diff
@@ -100,6 +102,7 @@ _expand_template = rule(
         "labels_to_flags": attr.string_dict(mandatory = True),  # string keys instead of label_keyed because Bazel doesn't support parsing wildcard target patterns (..., *, :all) in BUILD attributes.
         "exclude_external_sources": attr.bool(default = False),
         "exclude_headers": attr.string(values = ["all", "external", ""]), # "" needed only for compatibility with Bazel < 3.6.0
+        "enable_swift": attr.bool(default = False),
```
cpsauer, line 105: What do you think about just always having swift enabled, eliminating the need for configuration and branching?

I'm asking because swift extraction should be really fast, right, since there's no need for header extraction. Normal clangd doesn't crash or otherwise error out if there are swift commands, right?

You're the expert here, but it seems from the outside like this is worth having on all the time :)

refresh.template.py
```diff
@@ -93,6 +93,23 @@ def _print_header_finding_warning_once():
     Continuing gracefully...""")
 _print_header_finding_warning_once.has_logged = False
 
+def _file_exists(source_file, warn_missing_generated = True):
```
cpsauer, line 96: Thanks for your care in preserving this for swift :) It really does help a lot of people with generated code.

[Small tweak: If we're always leaving warn_missing_generated = True, what about just omitting the parameter, switching the return values, and naming it something like `_warn_if_file_doesnt_exist`]

refresh.template.py
```diff
@@ -694,7 +734,12 @@ def _apple_platform_patch(compile_args: typing.List[str]):
         # Undo Bazel's Apple platform compiler wrapping.
         # Bazel wraps the compiler as `external/local_config_cc/wrapped_clang` and exports that wrapped compiler in the proto. However, we need a clang call that clangd can introspect. (See notes in "how clangd uses compile_commands.json" in ImplementationReadme.md for more.)
         # Removing the wrapper is also important because Bazel's Xcode (but not CommandLineTools) wrapper crashes if you don't specify particular environment variables (replaced below). We'd need the wrapper to be invokable by clangd's --query-driver if we didn't remove the wrapper.
-        compile_args[0] = _get_apple_active_clang()
+        # Bazel wrapps the swiftc as `external/build_bazel_rules_swift/tools/worker/worker swiftc ` and worker has been removed in apple_swift_patch
+
+        if compile_args[0][-6:] == "swiftc":
```
cpsauer, line 739: Nit: maybe endswith would be nicer?

refresh.template.py
```diff
@@ -683,7 +694,36 @@ def _get_apple_active_clang():
     # Unless xcode-select has been invoked (like for a beta) we'd expect, e.g., '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang' or '/Library/Developer/CommandLineTools/usr/bin/clang'.
 
 
-def _apple_platform_patch(compile_args: typing.List[str]):
+@functools.lru_cache(maxsize=None)
+def _get_apple_active_swiftc():
```
cpsauer, line 698: Just checking my understanding: Does sourcekit-lsp have problems if we don't unwrap the main executable? I (I'm not sure if it's ever executed, like with query-driver in clangd--or how tolerant they are of unknown flags.)

refresh.template.py
```diff
@@ -761,6 +806,40 @@ def _get_cpp_command_for_files(compile_action):
...
+
+    Compile_action was produced by rule_swift (https://github.com/bazelbuild/rules_swift)
+    """
+    # Patch command
+    compile_action.arguments = _all_platform_patch(compile_action.arguments)
+    compile_action.arguments = _apple_swift_patch(compile_action.arguments)
+    compile_action.arguments = _apple_platform_patch(compile_action.arguments, compile_action.environmentVariables)
+
+    # Source files is end with `.swift`
+    source_files = set(filter(lambda x: x[-6:] == ".swift", compile_action.arguments))
```
cpsauer, line 820: Do we expect multiple--or files without a swift extension?
(Another good spot for endswith if you want.)

refresh.template.py
```diff
@@ -683,7 +694,36 @@ def _get_apple_active_clang():
...
+    compile_args.pop(0)
+
+    # Expand -debug-prefix-pwd-is-dot
+    match = next((i for i,v in enumerate(compile_args) if v == "-Xwrapped-swift=-debug-prefix-pwd-is-dot"), None)
+    if match:
+        compile_args[match] = "-debug-prefix-map"
+        compile_args.insert(match + 1, os.environ["BUILD_WORKSPACE_DIRECTORY"] + "=.")
+
+    # Remove other -Xwrapped-swift arguments like `-ephemeral-module-cache` `-global-index-store-import-path`
+    # We could override index-store by config sourcekit-lsp
```
cpsauer, line 720: What's our plan for integrating their indexing during build?
(Eventually we'll need to document how to configure things in the README, but that can wait until the end.)

refresh.template.py
```diff
@@ -683,7 +694,36 @@ def _get_apple_active_clang():
     # Unless xcode-select has been invoked (like for a beta) we'd expect, e.g., '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang' or '/Library/Developer/CommandLineTools/usr/bin/clang'.
 
 
-def _apple_platform_patch(compile_args: typing.List[str]):
+@functools.lru_cache(maxsize=None)
+def _get_apple_active_swiftc():
+    """Get path to xcode-select'd swiftc version."""
+    return subprocess.check_output(
+        ('xcrun', '--find', 'swiftc'),
+        stderr=subprocess.DEVNULL, # Suppress superfluous error messages like "Requested but did not find extension point with identifier..."
```
cpsauer, line 702: It looks to me from a quick test like this has been fixed in the latest Xcode.

That is, `xcrun --find clang --no-cache` and `xcrun --find swiftc --no-cache` now run clean--and I really shouldn't be blanket ignoring error messages needlessly. Would you mind removing stderr>devnull here and for clang? (As a favor. My issue, but good to keep the two in sync.)

refresh.template.py
```diff
@@ -683,7 +694,36 @@ def _get_apple_active_clang():
...
+        encoding=locale.getpreferredencoding()
+    ).rstrip()
+
+
+def _apple_swift_patch(compile_args: typing.List[str]):
+    """De-Bazel(rule_swift) the command into something sourcekit-lsp can parse."""
+
+    # Remove build_bazel_rules_swift/tools/worker/worker
+    compile_args.pop(0)
```
cpsauer, line 711: Could you explain/doc a little bit more about what's happening here and below? I assume these are things built into the wrapper scripts, maybe with a persistent worker? (Sorry for my ignorance--I just haven't worked much with Bazel&swift)

refresh.template.py
```diff
@@ -703,7 +748,7 @@ def _apple_platform_patch(compile_args: typing.List[str]):
         # We also have to manually figure out the values of SDKROOT and DEVELOPER_DIR, since they're missing from the environment variables Bazel provides.
         # Filed Bazel issue about the missing environment variables: https://github.com/bazelbuild/bazel/issues/12852
         compile_args = [arg.replace('__BAZEL_XCODE_DEVELOPER_DIR__', _get_apple_DEVELOPER_DIR()) for arg in compile_args]
-        apple_platform = _get_apple_platform(compile_args)
+        apple_platform = _get_apple_platform(compile_args, environmentVariables)
```
cpsauer, line 751: Ooh, are the environment variables specified correctly for the swift toolchain? If so, this'd be interesting to report on the issue linked above, since they'd been asking about whether it was cc specific.

cpsauer (commented):
@xinzhengzhang, thanks so much for all your work here, for taking the lead, and for simplifying, unifying the implementations, and posting this. And sorry again for my holiday slowness. You can expect me to be faster in the future.

I'll comment on things that jump out to me as being worth discussing in this first round. Thanks for being great!

refresh.template.py
```diff
@@ -683,7 +694,36 @@ def _get_apple_active_clang():
...
+    """De-Bazel(rule_swift) the command into something sourcekit-lsp can parse."""
+
+    # Remove build_bazel_rules_swift/tools/worker/worker
+    compile_args.pop(0)
+
+    # Expand -debug-prefix-pwd-is-dot
+    match = next((i for i,v in enumerate(compile_args) if v == "-Xwrapped-swift=-debug-prefix-pwd-is-dot"), None)
+    if match:
+        compile_args[match] = "-debug-prefix-map"
+        compile_args.insert(match + 1, os.environ["BUILD_WORKSPACE_DIRECTORY"] + "=.")
```
cpsauer, line 717: Like with clang, might be able to just strip this out, since the commands are conceptional happening in place, not that we're actually running/debugging with them anyway.

cpsauer:
Separate question: Do you know how source kit lsp gets its clangd?
(Usually we drop workarounds when clangd has a new release--I want to make sure we can keep doing that.)

xinzhengzhang:
> Separate question: Do you know how source kit lsp gets its clangd?
> (Usually we drop workarounds when clangd has a new release--I want to make sure we can keep doing that.)

clangd was found by searching the bin path
https://github.com/apple/sourcekit-lsp/blob/main/Sources/SKCore/Toolchain.swift#L123

refresh_compile_commands.bzl
xinzhengzhang: This flag designed for some other tools like [infer](https://fbinfer.com)
In bilibili we are using it for static analysis and it didn't support swift yet...

refresh.template.py
xinzhengzhang: This is a compile command argument sample and it seems that the only way to identify the source is ending with .swift

```
"arguments": [
      "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swiftc",
      "-target",
      "arm64-apple-macos16.1",
      "-sdk",
      "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk",
      "-emit-object",
      "-output-file-map",
      "bazel-out/darwin_arm64-dbg/bin/app/home/home.output_file_map.json",
      "-emit-module-path",
      "bazel-out/darwin_arm64-dbg/bin/app/home/app_home_home.swiftmodule",
      "-enable-bare-slash-regex",
      "-DDEBUG",
      "-Onone",
      "-Xfrontend",
      "-serialize-debugging-options",
      "-enable-testing",
      "-g",
      "-debug-prefix-map",
      "/Users/xinzhengzhang/Workspace/rules_gripper/examples/iosapps=.",
      "-Ibazel-out/darwin_arm64-dbg/bin/services/app/home",
      "-Ibazel-out/darwin_arm64-dbg/bin/services/base/router",
      "-Ibazel-out/darwin_arm64-dbg/bin/services/app/live/liveService",
      "-Ibazel-out/darwin_arm64-dbg/bin/external/build_bazel_rules_gripper/GPModule",
      "-Xcc",
      "-iquote.",
      "-Xcc",
      "-iquotebazel-out/darwin_arm64-dbg/bin",
      "-Xfrontend",
      "-color-diagnostics",
      "-enable-batch-mode",
      "-module-name",
      "app_home_home",
      "-parse-as-library",
      "-static",
      "-Xcc",
      "-O0",
      "-Xcc",
      "-DDEBUG=1",
      "-Xcc",
      "-fstack-protector",
      "-Xcc",
      "-fstack-protector-all",
      "app/home/home.swift"
    ],
```

refresh.template.py
xinzhengzhang: ok

refresh.template.py
xinzhengzhang: I tried and it does seem to work without unpacking the main executable in sourcekit-lsp but it can not tolerant other unknown flags like `-Xwrapped-swift`

`error: unknown argument: \'-Xwrapped-swift=-debug-prefix-pwd-is-dot\'error: unknown argument: \'-Xwrapped-swift=-ephemeral-module-cache\'"))
  Code: -32001 `

And it is still necessary for c&cpp because we need to extract header information.

To be on the safe side, I think it's better to unwrap it

refresh.template.py
xinzhengzhang: We can analyze commands in `aquery` and pre build .swiftmodule by striping `-emit-object`  in swift but that looks like it's going to be the second super slow header extractor...

For some small project fully build the target and cache the results by bazel ecosystem seems the best way but for some super huge project like bilibili it is impossible to wait a fully build before writing code.  So I am doing another set of rules and a vscode plugin for integrating indexing with building.  I will sync that once there is any progress

refresh.template.py
xinzhengzhang: ok

refresh.template.py
xinzhengzhang: Of course correct, the environment is resolved by the apple_common provider in 

https://bazel.build/rules/lib/apple_common#apple_host_system_env
https://bazel.build/rules/lib/apple_common#target_apple_env

refresh.template.py
xinzhengzhang: Done

https://github.com/hedronvision/bazel-compile-commands-extractor/commit/4fbd4a38987aae6801b637dac43a88729b951b51

refresh.template.py
xinzhengzhang: https://github.com/hedronvision/bazel-compile-commands-extractor/commit/f9ceaf3b7bf5579d7b1793571e5dafe35e4621e2

refresh.template.py
xinzhengzhang: https://github.com/hedronvision/bazel-compile-commands-extractor/commit/c7a4df6b6854f5d9bbaeb163b04eeac8b5695454

refresh_compile_commands.bzl
cpsauer: Ah! Got it. And to confirm, the static analysis tools crash, rather than gracefully ignoring the swift files? Bummer.

cpsauer: How about it we at least turn swift on by default?
Perhaps "exclude_swift" defaulting to False?
(We should make sure to eventually document this in the README and file docstring)

cpsauer: It might also be worth submitting bugs to the tools that crash on swift files? They're going to see more and more of those inputs as the years roll on, I think.

refresh.template.py
cpsauer: Ah, sorry. The `--no-cache` was just for testing--and would have exposed the superfluous error bug if Apple hadn't fixed it.  Thanks for fixing! (And, I'm realizing, giving me commit access. I should have just done.)

refresh_compile_commands.bzl
xinzhengzhang: https://github.com/xinzhengzhang/bazel-compile-commands-extractor/commit/22317ce83a6dd76673efc11ac91b39630d8cb135

cpsauer: Sounds good. But if it removes the need for the file filter, I'd still be happy with keeping an `exclude_swift` flag (off by default) for you, until those tools become tolerant to swift files?
(They do crash, yes, or otherwise issue errors that interfere with their use?)

refresh.template.py
cpsauer: Main executable: After some more thought, I think we should just set it to 'swiftc', rather than finding the full path via xcrun--assuming that works. 
Why? This'll keep things more open for Swift on other platforms down the line that don't have xcrun (like Linux, Swift for servers). If a tool does try to resolve swiftc, it should work just fine without xcrun, resolving to the same system default via the path, so we're stripping out unnecessary complexity. This works for clang, too, so I've removed the xcrun call for clang on the main branch in a797a787e56fb766c8a900c3ed5a650ea45be844. Sorry for leading you down an unnecessarily complex path to start; thanks for helping make this tool better all around.

Flags: Just to make sure I'm on the same page, these are red-squiggly errors in the editor, right? Makes total sense that we'll need to remove/replace them, then.

refresh.template.py
cpsauer: For clarity: I wasn't meaning to propose that we force a full swift build up front when it isn't needed, but we should help people get sourcekit-lsp configured correctly, via instructions in the readme.

I'm guessing part of that will be pointing source kit-lsp to pick up the indexing output generated during bazel builds? And configuring bazel builds to generate that output if they don't already? 

I'd love to learn more about what you're working on with the plugin!

refresh.template.py
cpsauer: Those flags look handy for configuring sourcekit-lsp! (like in https://github.com/hedronvision/bazel-compile-commands-extractor/pull/89#discussion_r1037779449)

But I just meant that it might make sense just omit that particular flag, -debug-prefix-map, since I don't think it'll impact analysis--nor compilation if it's just the cwd.

refresh.template.py
cpsauer: Got it. Is there a case we need to worry about where the user disables persistent workers?

xinzhengzhang: At least in the current rules_apple version we don't need. 
The [worker](https://github.com/bazelbuild/rules_swift/blob/master/tools/worker/worker_main.cc) has wrapped both.

refresh.template.py
xinzhengzhang: Here we should set `clang` in else branch or just assert the main executable has to be one of ['swiftc' , 'external/local_config_cc/wrapped_clang'] ?

refresh.template.py
xinzhengzhang: [Here](https://github.com/xinzhengzhang/bis/blob/main/refresh_compile_commands.bzl) is my implementation.
I have wrapped the `hedron_refresh_compile_commands` and collect all the module info which depended by the target into outputs.
Not as painful as rebuilding .d file for find headers, we can share module objects with compiling and indexing in this way.

refresh.template.py
xinzhengzhang: Done
https://github.com/xinzhengzhang/bazel-compile-commands-extractor/commit/e96bbbee80feee3425e7fb6e510bc20d430619be

refresh_compile_commands.bzl
xinzhengzhang: Done 
https://github.com/xinzhengzhang/bazel-compile-commands-extractor/commit/047daf96b94fd1cb46453336a9dee5f9d74a023a

xinzhengzhang:
Hi Chris~
Is there any other do I need to do about this patch?

cpsauer:
Hey @xinzhengzhang!

Could I ask you to take one more pass through, seeing if there are further ways to simplify the refresh.template.py changes and generally improving things in any way you see? 

I'm wondering if it might simplify things to unify the swift code path further with the cc codepath, just conditioning on action.mnemonic when needed rather than routing--or if they're actually different enough that they should be more fully separate. I'm not sure how many args patches are really shared. It'll be harder for C++ users to update things if the codepaths are too unified, but if they should share most logic, then we want to unify them so swift gets more future improvements. Your call, but I think we should definitely go more to one extreme or the other or at least document things so that we're clear. Please let me know what you think.

Let's also rename `_apple_swift_patch` -> `_swift_patch`, if that's OK, since swift may well get use on other platforms eventually. More generally, I'm not sure we can rely on __BAZEL_XCODE in the args, and maybe we can avoid enforcing an ordering on the patches? Perhaps we just only swap in clang if it's not swift--and otherwise rely on the popping logic in `_swift_patch`?

We're also going to at least need some docs to merge. If you want to just outline them, I can flesh them out, since I think I've got comparative advantage on English. (Not meaning any offense; your English is far better than I would be in any other language.) In particular, I think it'd be good to get instructions for pointing sourcekit-lsp into the accumulating index from building.

I can dive in more tomorrow, but it's getting late enough here that I'd better go to sleep for tonight.

Cheers,
Chris

xinzhengzhang:
Sorry, maybe I misunderstood before https://github.com/hedronvision/bazel-compile-commands-extractor/issues/1#issuecomment-1324776167 , I thought you wanted to make commands_extractor more general and adapt to more languages.
So I chose to separate the implementation of cc and swift directly from the entrance. However, limited by the structure of the current implementation of rules, I did not split it into multiple python files to make it more modular, but directly differentiated it in refresh.template.py.
Indeed, it looks like neither side is relying on it now...

Honestly, there is not much difference between the two implementations except where to find the source files. I will try to write another version and try not to break the original implementation of refresh.template.py.

cpsauer:
Ah, no sorry. Miscommunication somewhere. (About to go to sleep but want to catch you fast so I don't waste your time.)

cpsauer:
To make sure we're on the same page: Let's keep things in one file. Definitely looking to generalize things!

I was just wondering if maybe the codepaths could benefit from being merged further. (Truly wondering; unsure how much of the cc flag patching applies to swift.)

xinzhengzhang:
https://github.com/hedronvision/bazel-compile-commands-extractor/blob/c6cd079bef5836293ca18e55aac6ef05134c3a9d/refresh.template.py#L563

But there are indeed multiple sources in one swift compile action. 
And I also confirmed again that [features](https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/feature_names.bzl) in rules_swift can't be split into compiled single action.
Can I extend here to return multiple sources?

xinzhengzhang:
Because it was re-implemented, I re-opened a PR #96  for the convenience of review

PR #104, opened 2023-02-15
Use shell to resolve `bazel`

alejandrofsGR:
Otherwise user shell aliases are not visible in `subprocess`. This helps in systems where `bazel` might not be directly in the PATH, but rather aliased to a specific location.

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -884,6 +884,7 @@ def _get_commands(target: str, flags: str):
         capture_output=True,
         encoding=locale.getpreferredencoding(),
         check=False, # We explicitly ignore errors from `bazel aquery` and carry on.
+        shell=True, # Resolve `bazel` using the shell. Otherwise user aliases are not visible in subprocess.
     )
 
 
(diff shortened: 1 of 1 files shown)


cpsauer:
Hey @alejandrofsGR! Thanks for using the tool and looking to make it better than you found it.

I hear you on the need--but see that you quickly closed this. For my learning, can I ask what made you change your mind?

alejandrofsGR:
Hi Christopher. Sure, thanks to you for writing this cool tool! The reason I closed the PR is because I later found out setting `shell=True` seems to change how the list of arguments is interpreted by `subprocess.run(...)`. With `shell=True`, arguments seemed to be resolved as individual executables and I was getting errors like `aquery: command not found`. I tried to work around it by joining all arguments into a single string but all in all it was getting complicated so I decided to just put `bazel` in the path and remove the alias.

cpsauer:
Got it! Thanks, man. Appreciate it. (Helps me understand in case this comes back up.) I think one other is that shell=True is going to toss you into /bin/sh, which may or may not be configured with the aliases you want.

Anyway, thanks so much for your eagerness to contribute! I appreciate your taking the time to toss up a PR, try to make things better, and then explain.

Happy coding!
Chris

PR #113, opened 2023-03-18
Use pre-defined arguments to support override symlink prefix and append targets

LoSealL:
## Problem statement

Currently `bazel-compile-commands-extractor` works very well in most cases, but there are still some rare cases where users may specify a different output directory other than the default `"bazel-*"`.
On the other hand, if users add some new source files and want clangd to index them, they probably do not want to re-run the whole extractor, especially when users add new files frequently (usually at the beginning of a project or adding new features to an existing project), just because a total re-run takes quite a few miniutes.

## Solution
Hence we propose a PR to solve above two problems:
We add explicit argument parser to parse two pre-defined arguments: `"-b", "--bazel-out"` and `"-t", "--targets"`:
   - Override symlink: bazel run @hedron_compile_commands//:refresh_all -- --bazel-out=mypredix-
   - Append targets: Only parser //foo:bar target and append to existing compile commands. bazel run @hedron_compile_commands//:refresh_all -- -t "//foo:bar"

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -19,7 +19,7 @@ if sys.version_info < (3,6):
     # When adding things could be cleaner if we had a higher minimum version, please add a comment with MIN_PY=3.<v>.
     # Similarly, when upgrading, please search for that MIN_PY= tag.
 
-
+import argparse
 import concurrent.futures
 import enum
 import functools  # MIN_PY=3.9: Replace `functools.lru_cache(maxsize=None)` with `functools.cache`.
@@ -830,12 +830,12 @@ def _convert_compile_commands(aquery_output):
             }
 
 
-def _get_commands(target: str, flags: str):
+def _get_commands(target: str, flags: str, extra_flags: str=''):
     """Yields compile_commands.json entries for a given target and flags, gracefully tolerating errors."""
     # Log clear completion messages
     log_info(f">>> Analyzing commands used in {target}")
 
-    additional_flags = shlex.split(flags) + sys.argv[1:]
+    additional_flags = shlex.split(flags) + extra_flags
 
     # Detect anything that looks like a build target in the flags, and issue a warning.
     # Note that positional arguments after -- are all interpreted as target patterns. (If it's at the end, then no worries.)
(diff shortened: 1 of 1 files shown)


cpsauer:
Hello again, @LoSealL! Great to hear from you.

I hear you on both needs.

**On the faster, incremental update:**

I totally get that for very large codebases, rerunning the whole thing is too slow. (There are the usual speed tricks in the README, but it sounds like those are insufficient.)

There's a trap if you do it per-sub-target, though, and that's that the top-level-target often configures the targets it depends on, meaning that the commands for the sub targets are wrong if you ask for them the obvious way 🤦🏻. Cross-compiling, e.g. to mobile or embedded, is a key example of this. You'll get commands for the host platform, not the target platform. Plus, getting all dependencies of a sub-target can still be overwhelming.

So instead, I've been working with @xinzhengzhang for a while on a `--file` flag that lets an editor quickly ask for the correctly configured flag for a single file. The idea is that the editor would be configured to auto-call this tool to get the right command when files are opened or created, keeping things up to date, just in time.

Does that sound like it'd solve things for you too?

If so, we could really use your help, double checking that it works for you and resolving some of the remaining TODOs in the code. (Am I right in remembering that you'd helped with windows before? If so, there are some bite-sized windows TODOs we could really use your help on.) 

The work is going on in #99 with the early design discussion in #93. If you're down to help, I'm hoping he can just give you commit access the branch we're using for the feature. 

**For Bazel-out:**
I hear you--it'd be nice to handle all combos of bazel flags. But this one is significantly trickier than you might think, since you also have to port over all the bazel flags that reference bazel-out. There have been some other efforts at this and at adding an option to remove the external symlink (run a quick search through PRs and issues), but I really think it's more hassle than it's worth it, since it makes things much simpler to have a workspace that mirrors the compilation sandbox. That said, I remain open to it if you want to do the work.

Cheers--and thanks for being the kind of person that looks for ways to help improve things!
Chris

LoSealL:
Thanks @cpsauer, indeed it will be more convenient to interact with editor automatically, I will study how #99 is designed.

BR

cpsauer:
Thanks for taking a look, and being down to merge efforts @LoSealL. Please let me know what you think.

And for my learning, what does BR stand for?

LoSealL:
Best regards :-)

PR #125, opened 2023-05-31
Support ccache

manshengyang:
My project uses ccache not installed at the standard location and this leads to weird behaviors including generating all .o files in the workspace directory and clangd not able to find the correct gcc include paths.
The patch detects ccache usage and replace it with the actual compiler path.

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -760,6 +760,14 @@ def _all_platform_patch(compile_args: typing.List[str]):
             new_compile_args.append(arg)
     compile_args = new_compile_args
 
+    # replace ccache with the actual compiler
+    compiler_path = os.readlink(compile_args[0])
+    if compiler_path.endswith("ccache"):
+        compiler = os.path.basename(compile_args[0])
+        real_compiler_path = shutil.which(compiler)
+        if real_compiler_path:
+            compile_args[0] = real_compiler_path
+
     # Any other general fixes would go here...
 
     return compile_args
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -760,13 +760,17 @@ def _all_platform_patch(compile_args: typing.List[str]):
             new_compile_args.append(arg)
     compile_args = new_compile_args
 
-    # replace ccache with the actual compiler
-    compiler_path = os.readlink(compile_args[0])
-    if compiler_path.endswith("ccache"):
-        compiler = os.path.basename(compile_args[0])
-        real_compiler_path = shutil.which(compiler)
-        if real_compiler_path:
-            compile_args[0] = real_compiler_path
+    # Discover compilers that are actually symlinks to ccache--and replace them with the underlying compiler
+    try:
+        compiler_path = os.readlink(compile_args[0])
+    except OSError:
+        pass
+    else:
+        if compiler_path.endswith("ccache"):
+            compiler = os.path.basename(compile_args[0])
+            real_compiler_path = shutil.which(compiler)
+            if real_compiler_path:
+                compile_args[0] = real_compiler_path
 
     # Any other general fixes would go here...
 
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -761,12 +761,9 @@ def _all_platform_patch(compile_args: typing.List[str]):
     compile_args = new_compile_args
 
     # Discover compilers that are actually symlinks to ccache--and replace them with the underlying compiler
-    try:
+    if os.path.islink(compile_args[0]):
         compiler_path = os.readlink(compile_args[0])
-    except OSError:
-        pass
-    else:
-        if compiler_path.endswith("ccache"):
+        if os.path.basename(compiler_path) == "ccache":
             compiler = os.path.basename(compile_args[0])
             real_compiler_path = shutil.which(compiler)
             if real_compiler_path:


cpsauer:
Hey, Mansheng! Good to meet you. Thanks for being the type of person who helps out and leaves things better for everyone. I really appreciate it.

You're also teaching me something new here. (Thank you!) I hadn't realized people used ccache with Bazel. I'm assuming it's already installed for other non-bazel things with the standard compiler paths symlinked to it, is that right?

cpsauer:
I was doing some quick pre-merge testing and made some quick changes so it wouldn't crash when the compiler isn't a symlink. (See commit, back linked above) NBD--just raising it because I figured it might be interesting learnings for you. I know I've tripped myself up over that one in the past.

cpsauer:
Before I merge: I'm noticing that I must have been wrong that these symlinks are installed in place of the normal compiler, because otherwise shutil.which would just return the ccache link again, right? (You did indeed say "not installed at the standard location" above, after all.)

If that's right, there are a few things I should ask first:
1) For my learning, if ccache isn't just installed system-wide for something else, what's the underlying motivation behind specially wiring up bazel to use ccache?
2) Any thoughts on how you'd generalize this to system-installed ccache? Perhaps doing the equivalent of `which -a` and then suppressing matches that are ccache symlinks?
3) It's surprising to me that ccache is littering the workspace with .o's and not responding to clangd's --query-driver detection of system includes. Assuming ccache is supposed to be a transparent compiler replacement, these seem like things that should really get fixed in ccache, obviating changes here.
What do you think? Should we pursuing this fix at the ccache layer instead? 
To try to help pierce down to the ccache layer: If the .o littering is happening while this tool is run, it's probably ccache not honoring the `--dependencies` (and lack of `-o` and other -M* and *-dependencies flags) that should suppress compilation and the output of .o files. ([clang docs](https://clang.llvm.org/docs/ClangCommandLineReference.html), if useful. We invoke the compiler with these flags from _get_headers_gcc to figure out which headers are used by which source files. The parallel flags for MSVC are `/showIncludes /EP`, added in _get_headers_msvc.) Clangd --query-driver pierces through wrappers with e.g. `gcc -E - -v`, so that'd be the thing to make sure ccache handles identically there.

A couple other smaller things, relevant if we continue down the path of patching this here rather than in ccache:
- You know more about how people install ccache than me: Do we expect people to have renamed ccache? Super minor, but it strikes me that if not `compiler_path.endswith("ccache")` should maybe just check for `basename == 'ccache'`, and if so perhaps we should just do `'ccache' in basename` otherwise to handle more renamings.
- FWIW, I did benchmark this quickly and it seems like the repeated os.readlink calls are plenty fast. (We've had to cache filesystem calls for speed elsewhere with `@functools.lru_cache(maxsize=None)`.) Might be worth your quickly testing the impact of shutil.which--or its replacement to suppress ccache symlinks?

Thanks again,
Chris

manshengyang:
Hi Chris,


> Before I merge: I'm noticing that I must have been wrong that these symlinks are installed in place of the normal compiler, because otherwise shutil.which would just return the ccache link again, right? (You did indeed say "not installed at the standard location" above, after all.)

The standard location for ccache symblinks are in `/usr/lib/ccache` (e.g. `/usr/lib/ccache/gcc`, and these would be 
 all pointing to `/usr/bin/ccache`) and this is normally not in PATH and people would configure their build system to use ccache, e.g. via `CC=/usr/lib/ccache/gcc`. I believe ccache internally also uses PATH to detect gcc because in the past I could simply change PATH to switch gcc version without changing any ccache config.

> 
> If that's right, there are a few things I should ask first:
> 
> 1. For my learning, if ccache isn't just installed system-wide for something else, what's the underlying motivation behind specially wiring up bazel to use ccache?
> 2. Any thoughts on how you'd generalize this to system-installed ccache? Perhaps doing the equivalent of `which -a` and then suppressing matches that are ccache symlinks?

I've just tested using the system's ccache, encountered the same issues, and successfully used the patch to get around them.

> 3. It's surprising to me that ccache is littering the workspace with .o's and not responding to clangd's --query-driver detection of system includes. Assuming ccache is supposed to be a transparent compiler replacement, these seem like things that should really get fixed in ccache, obviating changes here.
>    What do you think? Should we pursuing this fix at the ccache layer instead?
>    To try to help pierce down to the ccache layer: If the .o littering is happening while this tool is run, it's probably ccache not honoring the `--dependencies` (and lack of `-o` and other -M* and *-dependencies flags) that should suppress compilation and the output of .o files. ([clang docs](https://clang.llvm.org/docs/ClangCommandLineReference.html), if useful. We invoke the compiler with these flags from _get_headers_gcc to figure out which headers are used by which source files. The parallel flags for MSVC are `/showIncludes /EP`, added in _get_headers_msvc.) Clangd --query-driver pierces through wrappers with e.g. `gcc -E - -v`, so that'd be the thing to make sure ccache handles identically there.
> 

You are right, I agree that this seems to be a ccache bug but many people use ccache in their linux distro's repo and may not be able to get this fix, patching this here would make it more easily available.

> A couple other smaller things, relevant if we continue down the path of patching this here rather than in ccache:
> 
> * You know more about how people install ccache than me: Do we expect people to have renamed ccache? Super minor, but it strikes me that if not `compiler_path.endswith("ccache")` should maybe just check for `basename == 'ccache'`, and if so perhaps we should just do `'ccache' in basename` otherwise to handle more renamings.
> * FWIW, I did benchmark this quickly and it seems like the repeated os.readlink calls are plenty fast. (We've had to cache filesystem calls for speed elsewhere with `@functools.lru_cache(maxsize=None)`.) Might be worth your quickly testing the impact of shutil.which--or its replacement to suppress ccache symlinks?

Yes, it's more readable to use basename, I've updated the code.
Also tested the costs of the function calls used, if you think any of these is not fast enough, we can use `lru_cache`.
```
os.path.islink: ~1us
shutil.which: ~10us
```

> 
> Thanks again, Chris

Thanks, Mansheng

cpsauer:
:) Thanks for your reply, Mansheng.

Very interesting about the paths. To confirm, people don't link the to the ccache binary from `/usr/local/bin/<compiler_name>` (on standard PATH) like [in the manual](https://ccache.dev/manual/4.8.1.html#_run_modes), but rather from /usr/lib/ccache? And if so, [they don't put it on the path](https://askubuntu.com/questions/470545/how-do-i-set-up-ccache)? (Apologies if I've misread something somewhere. Or maybe the ccache'd-gcc-on-path case already works? I am just less experienced with ccache and want to make sure we don't miss a common case while we're at it.)

[I remain quite curious--for my learning--about what motivates wiring up bazel to use ccache! Any chance it'd work to just generate the compile commands without ccache, even if you'll do the final build with ccache?]

Could I ask you to also report the underlying bugs to ccache in parallel? Seems well worth ccache fixing even if we work around it.

Cheers and thanks. I appreciate your patience in helping me understand and making this robust.
Chris

manshengyang:
Hi Chris,

Sorry about the late reply.
> :) Thanks for your reply, Mansheng.
> 
> Very interesting about the paths. To confirm, people don't link the to the ccache binary from `/usr/local/bin/<compiler_name>` (on standard PATH) like [in the manual](https://ccache.dev/manual/4.8.1.html#_run_modes), but rather from /usr/lib/ccache? And if so, [they don't put it on the path](https://askubuntu.com/questions/470545/how-do-i-set-up-ccache)? (Apologies if I've misread something somewhere. Or maybe the ccache'd-gcc-on-path case already works? I am just less experienced with ccache and want to make sure we don't miss a common case while we're at it.)

Yes, I think most of the time people simply change the compiler to ccache in their build system (cmake, bazel, etc.). The setting-path approach forces all compilations to use ccache and is a bit intrusive. The `/usr/lib/ccache` is the default install location on many linux distro.
> 
> [I remain quite curious--for my learning--about what motivates wiring up bazel to use ccache! Any chance it'd work to just generate the compile commands without ccache, even if you'll do the final build with ccache?]

I found bazel's caching not as accurate as ccache. Sometimes I changed some configs or env variables that did not change the results of the compilation and bazel still rebuilt everything while ccache was able to reuse the cache.

Yes I think we could use a different build config (without ccache) to generate the compile commands, but why not make it simply work out of the box :)
> 
> Could I ask you to also report the underlying bugs to ccache in parallel? Seems well worth ccache fixing even if we work around it.
Sure, let me figure out how to reproduce this without bazel first and then report this to them.
> 
> Cheers and thanks. I appreciate your patience in helping me understand and making this robust. Chris

cpsauer:
Merging...now my turn to *really* apologize for being slow to reply.

Thanks again for being the type of person who helps out and leaves things better for everyone. I really appreciate it, and I hope I haven't permanently dissuaded you from contributing here. Know that I really appreciate it even if I dropped the ball this time; and we could really use your expert help if and when you run into anything else in the future.

Could I ask you to still report the underlying bug to ccache, if you haven't already, and drop a link back here so I can track?

Thanks again,
Chris

PR #142, opened 2023-10-05
Add bzl_srcs to support Stardoc

calebzulawski:


cpsauer:
Hey Caleb! Thanks for looking for ways to make this tool better. Sorry for being a slowpoke over here--I've gotten a bit buried with urgent stuff. 

Could I ask you to spell out for me a little bit more what need this solves for you and others? I'd been thinking we'd do manual docs here, probably, since the interface is so narrow, but I'm wondering if, perhaps, this is instead for documenting your usage of the tool for your team? 

Cheers and thanks for reaching out,
Chris

calebzulawski:
No problem!
It's not actually for documenting this library at all, though it might help with that too (I haven't tried).

We use stardoc to document our internal rules and the problem I encountered is that a couple of our bzl files `load` from this repo.  Stardoc requires all `load`s to be listed in the stardoc rule's `deps`, otherwise that file cannot be documented.  In other words, no bzl file that uses this repo can be documented without this fix.

We have our own internal rules that has its own workspace macro, which in turn loads this repo's workspace setup.  Adding `bzl_srcs` (which seems to be the convention used by many other rules) allows documenting our own internal `workspace.bzl`:

```python
stardoc(
    name = "workspace" ,
    input = "myworkspace.bzl",
    deps = ["@hedron_compile_commands//:bzl_srcs"],
)
```

cpsauer:
Got it! Thanks for taking the time to explain and enlighten, Caleb. Definitely want to unblock that and get you back on the mainline; sorry. I'll get this into our other repos, too.

I'll make some small changes (document, move into interface section, rename to bzl_srcs_for_stardoc, broaden glob) and merge immediately to save a round trip. But if there are changes you don't like or ways you think it could be better, please holler or toss up another quick PR and I'll merge straight away.

cpsauer:
Actually, Caleb, could I get your advice quickly in fixing this for other parallel repos--and in keeping this working here in the future? I presume one needs transitive deps, so for https://github.com/hedronvision/bazel-cc-filesystem-backport we'd need 
`filegroup(name = "bzl_srcs_for_stardoc", srcs = glob(["**/*.bzl"]) + ["@bazel_tools//:bzl_srcs"], visibility = ["//visibility:public"])` right? (and the same thing here, if and when deps arise, to cover the loads of http_archive and maybe?) Seems way broad, but it looked like the more granular versions were too visibility constrained.

cpsauer:
^ do really want to make sure we sync up on this

calebzulawski:
I'm not sure if there's a standard convention, but yes, stardoc needs all transitive deps as well so I think it's helpful to include `"@bazel_tools//:bzl_srcs"`.  I think in some situations I have needed to manually add dependencies, but that's inconvenient.  There's also [skylib's bzl_library](https://github.com/bazelbuild/bazel-skylib#bzl_library) which has a `deps` attribute, but I haven't used that--I think a filegroup is sufficient.  Thanks for merging this!

cpsauer:
Thanks so much for the contribution, both in code and knowledge!

cpsauer:
(have added to some other repos)

PR #152, opened 2023-12-21
Adds the ability to use absolute paths for bazel managed artifacts

csmulhern:
This allows compile_commands.json commands to work for users who use e.g. `--experimental_convenience_symlinks`.


cpsauer:
(Responding over at #151, since these are coupled.)
Cheers and again, thank you!
Chris

PR #161, opened 2024-01-24
Fix: exception when external link exists under windows and runfiles enabled

nosamad:
I have observed an exception when using compile commands under windows with runfiles enabled.
The error is, that a windows path object has no method removeprefix.

Steps to reproduce:
- Enable Windows Developer mode and runfiles

# Options in .bazelrc
build --enable_runfiles
startup --windows_enable_symlinks


cpsauer:
🤦🏻 This is my bug flipping things over to pathlib. Thanks so much for noticing and taking the time to fix.

Made a small tweak to compact things, since current_dest is already a pathlib.Path, and am merging. Please lmk if you have any further problems or if I've messed something up.

Thanks again for leaving things better than you found them :)

Happy coding!
Chris

PR #162, opened 2024-01-24
Fix: assertion when using zig toolchain

nosamad:
The zig toolchain uses the file extension .obj instead of .o

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -163,7 +163,7 @@ def _parse_headers_from_makefile_deps(d_file_content: str, source_path_for_sanit
     # For example, `d_file_content` might be: `"foo.o : foo.cc bar.h \\\n     baz.hpp"`.
     target, dependencies = d_file_content.split(':', 1)
     target = target.strip()  # Remove the optional trailing space.
-    assert target.endswith('.o'), "Something went wrong in makefile parsing to get headers. The target should be an object file. Output:\n" + d_file_content
+    assert target.endswith(('.o', '.obj')), "Something went wrong in makefile parsing to get headers. The target should be an object file. Output:\n" + d_file_content
     # Undo shell-like line wrapping because the newlines aren't eaten by shlex.join. Note also that it's the line wrapping is inconsistently generated across compilers and depends on the lengths of the filenames, so you can't just split on the escaped newlines.
     dependencies = dependencies.replace('\\\n', '')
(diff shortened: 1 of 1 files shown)


cpsauer:
Hello again @nosamad! Definitely down to merge. Couple quick questions:
1) For my learning, what changed your thinking vs last time? (I thought zig had been respecting Bazel's -o after all?) A windows case?
2) For the split, maybe we should split on `': '`, in case there are multiple absolute windows paths?

Thanks so much for working to leave things better than you found them :)
Chris

cpsauer:
^ Went ahead and did that--I'll merge, but would still love it if you'd reply when you get a chance!

nosamad:
> Hello again @nosamad! Definitely down to merge. Couple quick questions:
> 1) For my learning, what changed your thinking vs last time? (I thought zig had been respecting Bazel's -o after all?) A windows case?
I think it is a windows case. As far as I know, there is no option of defining the Artefakt File-Extension through a toolchain definition - but I didn't review what they do, so maybe there is a possibility.
> 2) For the split, maybe we should split on `': '`, in case there are multiple absolute windows paths?
I don't know if this approach would be sufficient, as dep-Files are valid Makefiles and theoretically every valid Makefile Expression is allowed in the context of targets and prerequisites of a rule - which also can be defined through the command line (gcc). Does a simple Makefile-Parser exist in python? On the other hand your "simple" approach works in most cases :)
> 
> Thanks so much for working to leave things better than you found them :)
Actually thank you for your nice work!
> Chris
Sorry for the late response.

nosamad:
> > Hello again @nosamad! Definitely down to merge. Couple quick questions:
> > 1) For my learning, what changed your thinking vs last time? (I thought zig had been respecting Bazel's -o after all?) A windows case?
> I think it is a windows case. As far as I know, there is no option of defining the Artefact File-Extension through a toolchain definition - but I didn't review what they do, so maybe there is a possibility.
> > 2) For the split, maybe we should split on `': '`, in case there are multiple absolute windows paths?
> I don't know if this approach would be sufficient, as dep-Files are valid Makefiles and theoretically every valid Makefile Expression is allowed in the context of targets and prerequisites of a rule - which also can be defined through the command line (gcc). Does a simple Makefile-Parser exist in python? On the other hand your "simple" approach works in most cases :)
> > 
> > Thanks so much for working to leave things better than you found them :)
> Actually thank you for your nice work!
> > Chris
> Sorry for the late response.

Sorry I was so in a hurry that I missed your main points. Are you calling the compiler for depfile creation? I will investigate further, but it will take some time. Maybe we should provide some integration tests?

Thanks

PR #188, opened 2024-05-16
Distinct name and repo_name

mmorel-35:
I created https://github.com/bazelbuild/bazel-central-registry/pull/1989/files

This makes the module name and repo_name distinct

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,4 +1,8 @@
-module(name = "hedron_compile_commands")
+module(
+    name = "bazel-compile-commands-extractor",
+    version = "0.0.0-dev",
+    repo_name = "hedron_compile_commands",
+)
 
 p = use_extension("//:workspace_setup.bzl", "hedron_compile_commands_extension")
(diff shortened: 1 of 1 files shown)


cpsauer:
Hi, Matthiew! Thanks for being the kind of person who tries to leave things better than you found them. Could I ask you to help me understand the underlying issue here? What are we trying to solve for by making the repo name different? (My best understanding is that [it's some kind of migration setting for folks that want to change names](https://bazel.build/rules/lib/globals/module#module)?)

cpsauer:
Hey, Matthieu! As in the other PR, would it work for your use case to use this tool as it's designed as a dev dependency

That is 
`bazel_dep(name = "hedron_compile_commands", dev_dependency = True)`
from the readme, rather than 
```
`bazel_dep(
     name = "hedron_compile_commands",
     version = "",
 )
```
in [your hessian2 PR](https://github.com/alibaba/hessian2-codec/pull/30/files).

(This tool is normally used directly as s dev dependency to make code editing and listing more ergonomic, and not brought in transitively. If you're using it for some other purpose, just lmk!)

PR #194, opened 2024-05-30
Allow overriding bazel executable

keith:
It's possible that users don't have a `bazel` in their path depending on your project configuration. This allows passing `--bazel whatever` to override to a custom wrapper, or anything else.


cpsauer:
(For any other readers: Centralizing discussion over in #191, since you'd expressed a preference for that one.)

keith:
closing for that one

PR #202, opened 2024-06-26
Support custom output configurations

chadnorvell:
This change adds two main features:

1. You can configure compile commands for certain targets to go into separate files

For embedded systems development in particular, `clangd` doesn't work well with `compile_commands.json` generated for multiple targets. For example, if you have a build target that runs on your host machine in one configuration and another that runs on device in another configuration, `compile_commands.json` will contain multiple conflicting compile commands for the same source files. `clangd` will just use the first one it finds, which may not be the one you want to use for code intelligence. It's convenient to have separate compile commands files for each target, and switch the file `clangd` uses depending on how you want to navigate your code.

By providing the `target_file_names` argument, you can associate targets with names. Separate compile commands files will be generated for each of the specified targets, and will be placed in subdirectories with the specified names. Compile commands for any targets that aren't specified in `target_file_names` will go into the main compile commands file, just like before.

2. You can specify a different output path

If you are generating multiple compile commands files, its preferable not to output them into the workspace root. So you can specify a separate output path, relative to the workspace root.

This patch doesn't change any existing behavior; if you don't add either of the new arguments to your invocation of `refresh_compile_commands`, everything will work exactly as it did before.

### DIFF
--- a/refresh.template.py
+++ b/refresh.template.py
@@ -1405,20 +1405,60 @@ def main():
         # End:   template filled by Bazel
     ]
 
-    compile_command_entries = []
-    for (target, flags) in target_flag_pairs:
-        compile_command_entries.extend(_get_commands(target, flags))
+    target_file_names = {
+        # Begin: template filled by Bazel
+        {target_file_names}
+        # End:   template filled by Bazel
+    }
 
-    if not compile_command_entries:
+    # Associates compilation database file names with lists of compile commands.
+    # __all__ is a special case: It's the "catch all" for any compile commands that aren't
+    # assigned to a specific file. If no targets are assigned to specific files (i.e., if
+    # target_file_names is empty), all compile commands will go into __all__ and end up
+    # in one file.
+    compile_command_sets = {
+        '__all__': []
+    }
+
+    for (target, flags) in target_flag_pairs:
+        # If the target has a specific file name assigned, put the compile commands in their
+        # own set, to be written to their own file.
+        if target in target_file_names:
+            target_name = target_file_names[target]
+            compile_command_sets[target_name] = list(_get_commands(target, flags))
+        # Otherwise, put them into the main file.
+        else:
+            compile_command_sets['__all__'].extend(_get_commands(target, flags))
+
+    if len(compile_command_sets) <= 1 and len(compile_command_sets['__all__']) == 0:
         log_error(""">>> Not (over)writing compile_commands.json, since no commands were extracted and an empty file is of no use.
     There should be actionable warnings, above, that led to this.""")
         sys.exit(1)
 
+    
+    if not (root_dir := pathlib.Path({out_dir})).exists():
+        root_dir.mkdir(parents=True)
+
     # Chain output into compile_commands.json
-    with open('compile_commands.json', 'w') as output_file:
-        json.dump(
-            compile_command_entries,
-            output_file,
-            indent=2, # Yay, human readability!
-            check_circular=False # For speed.
-        )
+    for target_name in compile_command_sets:
+        # If the target doesn't have a specified file name, put it into the "catch all"
+        # compilation database.
+        if target_name == '__all__':
+            file_path = root_dir / "compile_commands.json"
+        # Otherwise, write the database to the specific target file.
+        else:
+            target_dir = root_dir / target_name
+            target_dir.mkdir(exist_ok=True)
+            file_path = target_dir / "compile_commands.json"
+
+        # This condition is only relevant to __all__. If each specified target has a specified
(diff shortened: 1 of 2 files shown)


cpsauer:
Hey Chad! A good improvement. Thanks!

IMO the best solution here might be for clangd to properly handle multiple commands per file, like Xcode does. Filed an issue over there a few years ago: https://github.com/clangd/clangd/issues/681 that I think is the canonical one. Lmk if we might be able to get the Googlers over there to tackle. Last I was over there the development seemed to be pretty google-driven. Sam McCall, [`@sam-mccall`](https://github.com/sam-mccall) would be the guy to contact, I think.

(Need to run for the moment, but I'll get back to this soon.)

chadnorvell:
I agree it would be ideal for `clangd` to handle this, your issue describes the problem well. Beyond errors and suggestions, this impacts code navigation. In [Pigweed](pigweed.dev), we use a [facades/backend](https://pigweed.dev/docs/facades.html) architecture, and it's really useful to follow the source trail behind a particular interface and see what concrete code is executed on a particular platform. Or to see `constexpr`s evaluated in the editor with values that match the target platform (like memory alignment, data type sizes, etc.). In those cases, simply getting a union or intersection wouldn't be sufficient, and it's better to narrow down the scope of compile commands to a particular target.

cpsauer:
Hey, sorry, Chad. Most of the last 24h turned into taking care of a temporarily digestively ill family member 🥴 

Good points on the other browsing features. I tried to edit the clangd issue to better factor those in--but found myself thinking that one would probably want the union for other browsing features, too, right? That is, if there are multiple compile commands for different platforms, show both implementations for jump-to-definition, both ifdef branches for highlighting/analysis, both constexpr values, etc. If so, that's now reflected over there. If not, lmk so I can improve it further.

Of course, this tool should still support creating great, focused compile commands databases when people want it. Didn't mean to be implying otherwise--just wanted to quickly tap back, alerting that the best cross-platform development experience probably requires those fixes in clangd, and that you might want to engage Google resources there, too, since that might be on the critical path to a great developer experience. Otherwise clangd is just providing a single-platform view on cross-platform files, after all.

cpsauer:
Back to the main issue of supporting focused compilation databases: We'd settled into the same facades/backend pattern for cross-platform code, so I've got some firsthand understanding, but we solved things a little differently. So I want to make sure I'm fully understanding the need you're running into, so we get to the best implementation that meets your needs, too. I'm certainly open to this one, but I think it's worth a quick sync to make sure I don't have something even better for you and so that I understand your needs into the future.

In particular we were seeing: (1) code understanding/navigation for the wrong platform impeding editing (e.g. wrong ifdefs greyed out, wrong jump-to-definition through facade) (2) incomplete errors, especially when accidentally using platform-specific features (caught later by builds/tests). Does that sound the same? 

As clangd workarounds, we'd found ourselves: (a) listing the targets in priority order (i.e. host/test last) and (b) having multiple refresh_compile_command targets with different orderings (or more focused subsets) that we'd occasionally run when we were focusing on specific platforms (e.g. if we were focused on developing for "my_other_microcontroller" instead of "my_default_board"). When we'd rerun those other orderings, it'd replace the existing compilation database, and the changes would get hot reloaded by clangd and apply, with no additional config needed. The tool stayed fast because of its internal caching. And if we did end up browsing files for other platforms, during that session, well, they were still there further down in the ordering. 

I'm imagining that there's something that breaks that solution in your case--or that makes the multiple files necessary--and that I should probably understand. (Similarly, I'm imagining there's something motivating putting the dictionary in one target rather than the simpler design of having multiple targets, each specifying an output path.) I should say also that there's the case where a single target compiles a given file multiple times (like for a host tool that's also in the shipped binary.)

Happy also to hop on a GVC to quickly align if that'd be faster (or more fun!) than typing things out. You know how to reach me :)

Chris

chadnorvell:
I've been tinkering with this a bit and I think the new version is a better solution. We've been using it internally and it's been really useful!

Let's catch up on this next week.

ljishen:
Is there any update on this?

chadnorvell:
Hey @ljishen, we should be close to a resolution on this. I'll coordinate with @cpsauer this coming week. Thanks for the nudge!

ljishen:
> Hey @ljishen, we should be close to a resolution on this. I'll coordinate with @cpsauer this coming week. Thanks for the nudge!

I tried your patch and added the `out_dir` parameter, but I still get the same error if I build with `--symlink_prefix`
```
//bazel-out is missing. Please remove --symlink_prefix and --experimental_convenience_symlinks, so the workspace mirrors the compilation environment.
```

Issue #204, opened 2024-07-04
No source files found in compile args


kuntryn:
got this error when compiling abseil library. 
```
 in _get_files
    assert source_file_candidates, f"No source files found in compile args: {compile_action.arguments}.\nPlease file an issue with this information!"
[...]

PR #205, opened 2024-07-18
Add .h to the list of overrides.

eugeneo:
This is a lazy fix for #204 

In my case it prints a bunch of warnings of "-fsyntax-only" but generates usable `compile_commands.json`

This is how my version can be tested, add this to MODULE.bazel:
```
git_override(
    module_name = "hedron_compile_commands",
    remote = "https://github.com/eugeneo/bazel-compile-commands-extractor.git",
    commit = "844279a911d0a0adca480b299799f03f65290cce",
)
```

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -1,0 +1,110 @@
+{
+  "lockFileVersion": 11,
+  "registryFileHashes": {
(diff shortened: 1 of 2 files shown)


helly25:
I independently made an extened version of this in A modified version is available as https://github.com/hedronvision/bazel-compile-commands-extractor/pull/209 which also does not add `MODULE.bazel.lock`.

PR #215, opened 2024-08-30
Make the bazel command and threads count configurable

sthornington:
This is a quick thing we needed to do in our environment to get everything working.

1. Make the bazel command configurable - we need to call bazel from a wrapper script to set up some stuff, and as a consequence, the command that this tool subshells needs to use that too.  I made this as a --define command line parameter so that it could be passed from the original invocation of bazel when running the tool.

2. Make the max thread count configurable.  As a consequence of the above, we don't want many instances of our bazel wrapper running concurrently, that can cause spurious `bazel dump --action_cache` failures, so I added a rule parameter so we can just hard code it in our rule wrapper.

3. Almost forgot, I switched print_args from iostream to ANSI C using printf, so that building that doesn't have to know where/which C++ runtime etc to use.

This is my first time doing bazel so if these are the wrong approaches please let me know. Thanks!

### DIFF
--- a/print_args.cpp
+++ b/print_args.cpp
@@ -1,13 +1,13 @@
 // Prints the arguments passed to the script
 
-#include <iostream>
+#include <stdio.h>
 
 int main(int argc, char *argv[]) {
-  std::cout << "===HEDRON_COMPILE_COMMANDS_BEGIN_ARGS===\n";
-  for (int i = 1; i < argc; ++i) {
-    std::cout << argv[i] << "\n";
-  }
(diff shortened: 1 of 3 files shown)


sthornington:
In my private fork I have added:

    1. change print_args to use stdio instead of iostreams
    2. add max_threads rule parameter
    3. added BAZEL_COMMAND --define parameter
    4. switch to ld.gold so that building print_args doesn't barf on --threads
    5. new file extension exclusion list param to avoid using codegen files to compile headers in clang-tidy
    
Happy to update the PR with all of these if someone ever wants.

Issue #220, opened 2024-09-19
How to specific startup option?


neko-para:
Currently, there seems no way to apply options between `bazel` and `[cmd]`, only after `[cmd]` is available (e.g. `-- --config debug`).

My project has multiple bazelrc file, and I'd use `bazel --bazelrc=xxx.bazelrc build //:my_target` to build it. Thus refresh_compile_commands doesn't know this and failed to find the proper bazelrc.

neko-para:
Sorry, seems to be the dup of #170

Issue #170, opened 2024-02-18
Relevant arguments are lost by the `bazel aquery ...` portion of compile command generation


btalb:
I've had a setup evolve so that it is now using a `bazelrc` in a custom location (i.e. `bazel --bazelrc=/path/to/.bazelrc ...` is now how all of my commands are run).

The `bazel aquery ...` portion of the compile command generation is a raw call that doesn't receive any of the arguments forwarded to it. Some simple examples:
- `build --action_env=BAZEL_CXXOPTS="-std=c++20"` works when in a `bazelrc` in the default location, but not when in a custom location
- `--registry` args also go missing

It leaves me in a situation where my project builds perfectly fine, but I can't generate commands because the logic is in a non-standard `bazelrc` location or passed via the `bazel run ...` command itself.

I've:
- confirmed manually forcing those arguments in as a patch to `refresh.template.py` works, but that is to brittle to meet what I need
- looked for something dirty like recovering the arguments from the process tree, but the process holding the args isn't still around when the Python script runs
- tried messing around with the `refresh_compile_commands` function, but don't know enough about Bazel yet to know if it's even possible to access / forward these types of args

Any ideas?

btalb:
I've just seen the `-- args`  workaround, which functionally does the job.

It still doesn't work for `--bazelrc` as it treats it as a command arg rather than startup option (i.e. before the `aquery` command). 

I guess this is more a feature request then:
- low-hanging fruit: support startup options via the `--` arg passthrough (probably pulling the list from `bazel help startup_options` is the sanest path to go)
- is there any possible way to pull the args directly from the command (manually duplicating long arg lists / bazelrcs is tedious / error prone)?

cpsauer:
Hey, thanks for raising this, Ben.

I really wish we could pull the args from Bazel, but I don't know of a way. (Won't bazel help return the options from its invocation, which wouldn't automatically include the additional options)

Some workaround-ish ideas:
- Would it work to supply the basil RC option on both sides of the --?, that way passing it to both invocations?
- Alternately, would it work to import the bazelrc you want to use from one in the default location, thus sparing you having to pass it all the time?

Chris

btalb:
Thanks @cpsauer ; I had a feeling this may be the result.

- the passing `--bazelrc` both sides doesn't work because that's a [startup option](https://bazel.build/reference/command-line-reference#startup-options) which means it has to come before the `aquery` part of the command
[...]

PR #222, opened 2024-09-20
feat: support startup flags

neko-para:
Specific startup flags via an extra `--startup`.

```shell
$ bazel run //:refresh_compile_commands -- --startup --bazelrc=abc.bazelrc --config debug
```


neko-para:
Which should fix #220 #170

PR #235, opened 2024-12-17
Change py_binary to load from rules_python

stagnation:
To follow the (future) migration of the python rules out of the set of native rules: https://github.com/bazelbuild/bazel/issues/17773 . This requires us to load the 'rules_python' module and while at it we parameterize the 'py_binary' so it is possible for consumers to switch to their own macros, following the model of pip_compile: https://github.com/bazelbuild/rules_python/blob/9b2b70adba5431162401a97b2bbab1dc938e7245/python/private/pypi/pip_compile.bzl#L32

This solves loading errors:

    ERROR: Traceback (most recent call last):
        File "/home/nwirekli/.cache/bazel/_bazel_nwirekli/1eeb85843c9866d288d7d5644c3b4615/
        external/hedron_compile_commands+/refresh_compile_commands.bzl", line 67, column 27, in <toplevel>
                    py_binary = native.py_binary,
    Error: no native function or rule 'py_binary'
    Available attributes: ...
    ERROR: error loading package under directory '': error loading package '@@hedron_compile_commands+//':
    initialization of module 'refresh_compile_commands.bzl' failed


stagnation:
I think this is only needed because we use `8.0.0rc2`. Where the default value for a new flag was too strict. A colleague mentioned similar errors and that with the real 8.0.0 this is not a problem. So, after digging ourselves out of the module-version hole that prevents a clean upgrade to 8 this is helpful. And I hope to save some time in a later migration away from `native.py_binary`.

keith:
this will be required for bazel 9.x https://github.com/bazelbuild/bazel/commit/0522f8e58bf864005207eb8f09162d35ddd1f678, so probably want to merge this soon since that's ~2 months away

andrea-nisti:
any updates on this? currently we overcome this with a patch

stagnation:
I am not actively working on it, but it stands to reason that this, or something rebased should be merged, yes :)

PR #254, opened 2025-06-25
Fix for cases when aquery environment results do not contain a value key.

evanpurcell:
Aquery environment results may not always contain a 'value' field that are set to empty such as when setting a .bazelrc setting like the following:

`build --action_env=SOME_ENV=`

This should fix #252, please confirm.

PR #270, opened 2025-11-18
Fix no value environment variables in refresh.template.py

ImmortalEmperor:
Fix issue where environment variable has no value reported in bazel query.


ImmortalEmperor:
duplicate of #254

PR #272, opened 2025-12-12
Fixes for Bazel 9 and 10

theoparis:


theoparis:
See https://github.com/hedronvision/bazel-compile-commands-extractor/pull/281

